PR #3, opened 2015-04-12
GtkPixbuf: gdk_pixbuf_new_from_subpixbuf should be gdk_pixbuf_new_subpixbuf.

StephanvanSchaik:
This fixes a linking issue on Microsoft Windows, where it checks whether symbols exist during the linking phase, rather than run-time.


GuillaumeGomez:
Thanks for the fix of your code ! ;)

PR #5, opened 2015-04-14
Safety audit of `gdk::widgets::Pixbuf`

cybergeek94:
* Eliminates the possibility of use-after-free from passing the pixel pointer to `CVec`
    * Instead, returns a slice to the pixel bytes with a lifetime restricted to that of `&self`
* Added `pixels_rgba_mut()` convenience function for accessing a slice of pixels IFF the image is RGBA with 8 bits per pixel
    * Additional convenience functions could be added for different sets of channels (RGB, Grayscale, etc)
    * Could be polymorphic over the sample size (one function for all of `u8`, `u16`, etc)
* Updated `put_pixel` to use `pixels_rgba_mut()` and return early if it returns `None`
* Added `Drop` impl for `Pixbuf` that releases owned `GdkPixBuf` to prevent leaks
    * `gtk::widgets::Image` needs to be updated to increase the refcount to `GdkPixBuf` before creating this object in `get_pixbuf()`.
    * Alternately, the `impl_GObjectFunctions!` macro could increment the refcount in `wrap_pointer()`, but other objects would have to be updated accordingly.

[breaking change]

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,8 +19,6 @@ Bindings and wrappers for __GDK__
 
 */
 
-#![feature(unique)]
-
 extern crate gdk_sys as gdk_ffi;
 extern crate glib_sys as glib_ffi;
 extern crate glib as glib_main;
--- a/src/widgets/pixbuf.rs
+++ b/src/widgets/pixbuf.rs
@@ -17,9 +17,9 @@
 
 use glib::to_gboolean;
 use glib::translate::{FromGlibPtr, ToGlibPtr};
+use glib::ffi as glib_ffi;
 use ffi;
-use c_vec::CVec;
-use std::ptr::Unique;
+use std::slice;
 
 #[repr(C)]
 /// This is the main structure in the &gdk-pixbuf; library. It is used to represent images. It contains information about the image's pixel 
@@ -63,14 +63,15 @@ impl Pixbuf {
         unsafe { ffi::gdk_pixbuf_get_bits_per_sample(self.pointer as *const ffi::C_GdkPixbuf) }
     }
 
-    pub fn get_pixels_with_length(&self, length: &mut u32) -> Option<CVec<u8>> {
-        let tmp = unsafe { ffi::gdk_pixbuf_get_pixels_with_length(self.pointer as *const ffi::C_GdkPixbuf, length) };
+    pub fn get_bytes_mut(&self) -> Option<&mut [u8]> {
+        let mut length = 0u32;
+        let tmp = unsafe { ffi::gdk_pixbuf_get_pixels_with_length(self.pointer as *const ffi::C_GdkPixbuf, &mut length) };
 
         unsafe {
             if tmp.is_null() {
                 None
             } else {
-                Some(CVec::new(Unique::new(tmp), *length as usize))
+                Some(slice::from_raw_parts_mut(tmp, length as usize))
             }
         }
     }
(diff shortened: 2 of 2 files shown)


gkoz:
It would be a good occasion to try out the `glib::translate` system in practice, i.e. implement `FromGlibPtrNotNull<*mut ffi::C_GdkPixbuf>` for `Pixbuf` to encapsulate ownership management. Then the constructor could just use `FromGlibPtrNotNull::take()` and `Image` would do `FromGlibPtrNotNull::borrow()` and not fiddle with refcounting.

cybergeek94:
I was thinking something more along the lines of a wrapper for pointers that called `g_object_ref` on creation and `g_object_unref` on drop.

gkoz:
A generic wrapper would be even better but I don't think we're there yet. And you still need to support two ways of ownership transfer.

cybergeek94:
Well, it could have another function that skips `g_object_ref`. Both would be marked `unsafe`.

gkoz:
Or an `Object` could take care of reference counting once and let other wrappers defer to it. But we don't have this yet. We do have the way to express `NotNull` and `take`/`borrow` though.

cybergeek94:
The `take/borrow` convention makes it difficult to ensure an owned `PixBuf` does not leak memory without affecting the refcount of a borrowed `PixBuf`. I don't know if the pointer returned from `gtk_image_get_pixbuf()` is even *allowed* to increment the refcount, but it would be necessary if the `Drop` impl on `PixBuf` is to remain. A quick fix would be adding a `borrowed: bool` flag to the struct but it seems like it would be better if we didn't have to redo this for other types following this pattern.

I'm thinking of a wrapper for pointers to `C_GObject` types that expresses whether the pointer is owned or borrowed, and only does refcounting on owned:

```rust

// This type is deliberately non-copy
pub enum GObjectPtr<T> {
    Owned(*mut T),
    Borrowed(*mut T),
}

impl GObjectPtr<T> {
    pub fn from_owned(ptr: *mut T) -> GObjectPtr<T> {
        assert!(!ptr.is_null());
        GObjectPtr::Owned(ptr)
    }

    pub fn from_borrowed(ptr: *mut T) -> GObjectPtr<T> {
        assert!(!ptr.is_null());
        GObjectPtr::Borrowed(ptr)
    }

    pub fn as_ptr(&self) -> *mut T {
        match *self {
            GObjectPtr::Owned(ptr) => ptr,
            GObjectPtr::Borrowed(ptr) => ptr,
        }
    }

    pub fn as_gobject_ptr(&self) -> *mut C_GObject {
        self.as_ptr() as *mut C_GObject
    }

    pub unsafe fn to_owned(&self) -> GObjectPtr<T> {
        g_object_ref(self.as_gobject_ptr());
        GObjectPtr::Owned(self.as_ptr());
    }

    pub unsafe fn borrow(&self) -> GObjectPtr<T> {
        GObjectPtr::Borrowed(self.as_ptr())
    }
}

impl<T> Drop for GObjectPtr<T> {
    fn drop(&mut self) {
        if let GObjectPtr::Owned(ptr) = *self {
            unsafe { g_object_unref(ptr as *mut C_GObject); }
        }
    }
}
```
Later on, the `Borrowed` version can be extended to include `PhantomData<&'a T>` so Rust can ensure it won't outlive its source. This would necessitate adding a lifetime to every object containing a `GObjectPtr`, but objects that don't intend to have a borrowed version can use `GObjectPtr<'static, T>`.

We can restrict the type parameter with an unsafe trait that is only implemented by types that can be cast to `*mut C_GObject`:

```rust
pub unsafe trait AsGObject {}

// in gdk-sys
unsafe impl AsGObject for C_GdkPixBuf {}
```

And this way, memory management is encompassed in a single wrapper for all `C_GObject` types.

gkoz:
> I don't know if the pointer returned from gtk_image_get_pixbuf() is even allowed to increment the refcount

We need to find this out for certain.

> The caller of this function does not own a reference to the returned pixbuf.

This could just be an awkward way to say 'borrow'.

cybergeek94:
>>The caller of this function does not own a reference to the returned pixbuf.

>This could just be an awkward way to say 'borrow'.

I just took it to mean, "The refcount on this pointer is not incremented before it is returned, so don't decrement it or you'll screw things up." It says nothing about incrementing it ourselves.

gkoz:
Such a major breach of the reference counting pattern would be documented more explicitly and isn't the first logical thing to assume. The docs do sometimes talk in terms of ownership and not owning an object means you must increase the reference count of you want to hold on to it.

gkoz:
The reference counting management and returning of `CVec` have been taken care of in #24.

GuillaumeGomez:
Like said @gkoz, this PR is "useless" now (I can't find the word for it, sorry). Thanks for your work anyway ! I hope to see another contribution of you soon ! :-)

gkoz:
There was pixels_rgba_mut here also, but I can't evaluate its usefulness right now.
The hard question of how to allow safe access to the pixel data remains to be answered.

GuillaumeGomez:
I see it like this : using an opaque struct to make operations on it and then hand it back to gdk.

gkoz:
Opaque struct is okay because the pixbuf itself can be that struct. Letting the slice returned from `get_pixels` live on its own is unsafe. A holder of a `&[u8]` will depend on it not being mutated by anyone else and a holder of `&mut [u8]` will assume being the only one to have access to that memory. We can't guarantee neither.

PR #8, opened 2015-04-20
Support wrapping a screen pointer

gkoz:


GuillaumeGomez:
Why creating a new method ? The `wrap_pointer` isn't enough ?

gkoz:
You're right, my mistake

GuillaumeGomez:
Well, I merged and then reverted after seeing it. Just fix your gtk code.

PR #12, opened 2015-05-01
Add pixbufLoader and PixbufFormat

GuillaumeGomez:


src/widgets/pixbuf_format.rs
```diff
@@ -0,0 +1,124 @@
+// This file is part of rgtk.
```
gkoz: We're using another license header now

gdk-sys/src/lib.rs
```diff
@@ -626,4 +630,35 @@ extern "C" {
     // Gdk Key Handling                                                  NOT OK
     //=========================================================================
     pub fn gdk_keyval_name                         (keyval:c_uint) -> *mut c_char;
+
+    //=========================================================================
+    // GdkPixbufLoader                                                   NOT OK
```
gkoz: What does "not ok" mean?

GuillaumeGomez: It means so it hasn't been tested yet. A big part of rust-gnome is in the same case.

Issue #13, opened 2015-05-02
GErrors as supplementary returns (+ GQuarks and Error domains)


jashephe:
I recently encountered this thanks to the addition of `gdk::PixbufLoader` in #11.  `PixbufLoader`'s function `loader_write(&self, buf: &[u8], error: &mut Error)` requires passing a `GError` (in the form of `gdk::glib::Error`).  [In the GDK documentation](https://developer.gnome.org/gdk-pixbuf/unstable/GdkPixbufLoader.html#gdk-pixbuf-loader-write), the function is looking for a double pointer to memory for an error object, which presumably is only populated when `false` is returned.  In the Rust version, however, we have to pass a fully-instantiated `Error` struct, which can currently only be created with `new_literal(domain: u32, code: i32, message: &str)`.  However, because [`GQuark`](https://developer.gnome.org/glib/stable/glib-Quarks.html)s don't appear to be bridged yet, I cannot successfully create `Error`s because I have no way of determining the domain, and thus cannot write data to `PixbufLoader`s.

It seems to me that the real problem is that the original double-pointer function signature is not reproduced in a Rustic way.  Does anyone have any suggestions?  Thanks!

In the interest of full disclosure, I'm a beginner at all of this, so I might just be missing something obvious.

GuillaumeGomez:
It is reproduced but you just don't see it. Inside the function, we unwrap the pointer and its address to the function. However, it seems I should add a `Error::new()` without params. What do you think of that ?

gkoz:
@jashephe is right, that way doesn't work, we need to accept a new pointer to the error and take ownership of it.

GuillaumeGomez:
Hum... I'll look to do that later in the afternoon then.

jashephe:
Thanks a lot!  I'd love to try to get more involved and help out with this sort of thing, but I'm not really sure how this one should be implemented.

As @gkoz suggests, I think the ideal function signature would look something like 
```rust
fn loader_write(&self, buf: &[u8]) -> Result<(), Error>
```

GuillaumeGomez:
@jashephe: If you want you can try to do this one. You don't have anything to lose and we can help. :-)

jashephe:
Alright, I'll take a look.

jashephe:
Resolved by #15 and #16.

Issue #4, opened 2015-04-13
Improve null pointer handling


cybergeek94:
Many parts of this crate's API as well as some of the related ones return `Option` to handle when a pointer returned from an FFI call is null. Unless the C API documentation says why a returned pointer might be null, it is probably safe to assume that a null pointer means something bad happened and the program shouldn't continue (e.g. allocation failure/OOM). Overuse of `Option` for little good reason when they would usually prefer to unwrap it anyways pollutes client code and reduces ergonomics.

GuillaumeGomez:
I don't agree, a lot of gtk objects returns NULL even if it's not a fatal error. For example, if you want to open an inexistent file, you'll have a NULL pointer returned. And there are a lot of cases like this. If the object allocation function returns None, it's up to the user to see what and why it didn't work.

cybergeek94:
When an `Option` is `None` it doesn't imply any sort of error, just that the value requested is unavailable. In cases of the former kind it's pragmatic in Rust to return `Result`. Even if the return types for these methods were changed to `Result<_, ()>` until a more sane error type came forward, it would be a massive increase in ergonomics because it would work with the `try!()` macro, and it would make it obvious that the function can fail.

For many pointer-returning functions, though, the Gnome docs don't say if and/or when they could return null, which I would assume means a null pointer coming from these functions is exceptionally rare, such as an allocation failure. Forcing the user to treat these pointers as nullable might improve safety but at a massive cost to ergonomics and speed, because they're always taking the slow path. And panicking on a null pointer will tell them then and there that something went wrong.

Having `Option` everywhere is just noise, IMO.

GuillaumeGomez:
Basically, we had a `Result` returning stuff if I remember well. Why did we change ? I'd need to speak about that with @jeremyletang. Personally, I prefer the `Option` over the `Result`, but that's just my opinion...

I'm against panicking in libraries, except for the library initialization function (and yet, and I don't really like that). Why ? Because you prevent the developer to handle the library error and doing its own stuff to recover from it. For example, he uses gtk for something but can load dynamically another library to do the same thing if the first one fails. If the first llibrary panicks, he can't do anything.

cybergeek94:
[...]

PR #15, opened 2015-05-02
Fix PixbufLoader Error Returns

jashephe:
This appears to resolve the issue proposed in #13.  I also took the liberty of adding documentation (adapted from the official gdk-pixbuf docs), but if that isn't desired, I can pull that out.

However, in the process of doing this, I noticed that `new_with_type()` and `new_with_mime_type()` both have the same double-pointer error-as-return pattern.  Switching these over to the new style isn't as straightforward however, because there is also a possibility for failure if a `NULL` pointer is returned.  It seems that this is partially blocked by #4, and once a decision is made there, they can be changed.

### DIFF
--- a/src/widgets/pixbuf_loader.rs
+++ b/src/widgets/pixbuf_loader.rs
@@ -4,6 +4,7 @@
 
 //! GdkPixbufLoader — Application-driven progressive image loading.
 
+use std;
 use ffi;
 use glib::{to_bool, GlibContainer};
 use glib::translate::ToGlibPtr;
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
No no, that's a very good thing ! We have a serious lack on the doc point... We tried at first to provide a binding for the gnome libraries. Documentation wasn't a priority but it's very welcome ! Thanks a lot for your work !

jashephe:
Okay then, maybe I'll work on adding some more documentation when I have the time.

GuillaumeGomez:
> Right now I feel like just let mut error = ptr::null_mut(); in such cases would be preferable but I'm open to arguments against that.

I didn't pay attention but it would have been just as fine.

gkoz:
It seems we're in agreement about style so @jashephe would you mind doing a follow-up PR addressing my comments?

jashephe:
Would you like ::glib::Error switched to a global import as well?

gkoz:
Yes, we only have one Error type so far so there shouldn't be any confusion. Generally occurrences of leading "::" (except maybe when it's just one in the whole module) are undesirable, they're a consequence of rearranging the repository structure.

PR #16, opened 2015-05-03
Fix style in PixbufLoader

jashephe:
Resolves the changes discussed in #15.

### DIFF
--- a/src/widgets/pixbuf_loader.rs
+++ b/src/widgets/pixbuf_loader.rs
@@ -18,16 +18,12 @@ pub struct PixbufLoader {
 impl PixbufLoader {
     /// Creates a new pixbuf loader object.
     ///
-    /// # Failures
-    /// Returns `None` if the pixbuf loader cannot be created.
-    pub fn new() -> Option<PixbufLoader> {
+    /// # Panics
+    /// Fails if the pixbuf loader cannot be retrieved. 
+    pub fn new() -> PixbufLoader {
         let tmp = unsafe { ffi::gdk_pixbuf_loader_new() };
-
-        if tmp.is_null() {
-            None
-        } else {
-            Some(PixbufLoader::wrap_pointer(tmp))
-        }
+        assert!(!tmp.is_null());
+        PixbufLoader::wrap_pointer(tmp)
     }
 
     /// Creates a new pixbuf loader object that always attempts to parse image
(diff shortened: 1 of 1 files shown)


src/widgets/pixbuf_loader.rs
```diff
@@ -43,7 +43,7 @@ impl PixbufLoader {
     ///
     /// # Failures
     /// Returns `None` if the pixbuf loader cannot be created.
-    pub fn new_with_type(image_type: &str, error: &mut ::glib::Error) -> Option<PixbufLoader> {
+    pub fn new_with_type(image_type: &str, error: &mut Error) -> Option<PixbufLoader> {
```
gkoz: Oh, I didn't notice there were more functions using the same error handling pattern here. Care to make them all consistent?

jashephe: I mentioned that in #15:  https://github.com/rust-gnome/gdk/pull/15#issue-72717508
What are your thoughts?

gkoz: Right, I've missed that comment, sorry.
Generally, it should be okay to return `Err` in case `!error.is_null()` and `assert` that the return value is not null otherwise. And the gdk_pixbuf_loader seems to conforms to this model. The assertion is likely to be moved into a single place in the wrapper types refactoring later...

jashephe: Sounds good to me.  Pursuant to your comment in https://github.com/rust-gnome/gdk/issues/4#issuecomment-98530710, I will update the functions to match this style:
```rust
should_not_return_NULL_but_might_because_ffi_craziness() -> T // with the possibility of a panic!
might_validly_return_NULL() -> Option<T>
might_return_GError() -> Result<T, Error>
```
In this particular case, these will panic if `tmp` is `NULL`.

gkoz: Great, thanks!

src/widgets/pixbuf_loader.rs
```diff
@@ -18,16 +18,12 @@ pub struct PixbufLoader {
 impl PixbufLoader {
     /// Creates a new pixbuf loader object.
     ///
-    /// # Failures
-    /// Returns `None` if the pixbuf loader cannot be created.
-    pub fn new() -> Option<PixbufLoader> {
+    /// # Panics
+    /// Fails if the pixbuf loader cannot be retrieved. 
+    pub fn new() -> PixbufLoader {
```
GuillaumeGomez: We discussed about returning objects without Option wrapping somewhere but we didn't started it yet. Could it possible for now to keep returning `Option<PixbufLoader>` please ?

gkoz: Do you want that to make all breaking changes at once instead of gradually or some other reason?

GuillaumeGomez: All at once, or at least at once by library. If we do only a part at each time, we could forgot some and it would take a much longer time. So let's open an issue for each concerned library. After that, I'll put all rust-gnome libraries on crates.io (I'll do pango).

gkoz: This is tied up with returning `Result`s as well, so it seems you're saying don't change the APIs unless you're converting the whole crate?

GuillaumeGomez: No, from my point of view, `Result` isn't *directly* linked to returning `Option` types.

gkoz: `new_with_type` is a constructor that's being converted from `Option` to `Result`, sticking to `Option` means not doing it now.

GuillaumeGomez: For the `new_with_type`, it does make sense. For `new`, it's not necessary for the moment.

jashephe:
Okay, I've reverted the changes to `PixbufLoader::new()`, but left the other two `Result`-returning functions intact.

GuillaumeGomez:
Perfect ! Thanks !

src/widgets/pixbuf_loader.rs
```diff
@@ -6,7 +6,7 @@
 
 use std;
```
gkoz: It's more idiomatic to import `std::ptr` an then use `ptr::null_mut()`.

gkoz:
Beside a small nit, :+1:

GuillaumeGomez:
@gkoz: wo, you're really looking in detail ! However I agree, I find it more beautiful written like `ptr::null_mut()`.

@jashephe: Can you update it please ? :-)

jashephe:
@GuillaumeGomez Sure, I'll fix it when I get home tonight.

GuillaumeGomez:
Thanks !

GuillaumeGomez:
@jashephe: When you update, please notify me. Unfortunately, github doesn't send notification when PR is updated... Everything's fine, I merge. Thanks !

PR #21, opened 2015-05-09
Publish event members

lucaswerkmeister:


gkoz:
It would be nice to have some safe solution to ensure that the event record we receive in the signal handler is of the correct type (probably implement `FromGlibPtr` for events).

lucaswerkmeister:
I’m sorry, I don’t understand – is there some problem with this PR? Should some members remain unpublished?

gkoz:
These belong in `gdk-sys` actually, as `gdk` shouldn't expose the foreign types obviously. So a proper approach is to move them there (publishing the fields) and then most likely create wrapper structs with accessor methods that would take care of any conversions.

lucaswerkmeister:
Okay, but in the meantime, is there any reason why some events (e. g. `EventTouch`, `EventScroll`, `EventKey`) have public members, and others (e. g. `EventMotion`, `EventButton`) don’t? I assumed this was simply an oversight. AFAICT, there’s currently no way to get, for instance, the coordinates of a click event, and that seems… inconvenient :D

gkoz:
Can't answer for why some of the stuff has been done the way it is but it's way unfinished for sure.
I guess there's no harm in publishing scalar types. Do you want raw pointers as well?

lucaswerkmeister:
> Do you want raw pointers as well?

Not really… I’ll add a commit to unpublish them (including those that were public before).

gkoz:
What, the event structs are missing `#[repr(C)]`? :/
Please add that and it will allow you to remove `allow(dead_code)`.

src/events.rs
```diff
@@ -72,6 +72,8 @@ pub trait Event: Sized {
     }
 }
 
+// TODO unfisished, see some discussion in #21
```
gkoz: Make this #25

lucaswerkmeister: done (squash + rebase), also fixed the typo in “unfi*s*ished”

lucaswerkmeister:
Done.

gkoz:
:+1:

GuillaumeGomez:
Here I am ! @lucaswerkmeister: Nice job and thanks for your work !

lucaswerkmeister:
Cool, thanks for merging!

PR #23, opened 2015-05-09
Add missing functions from Pixbuf

GuillaumeGomez:


gkoz:
I'm about to push the refactored gdk...

GuillaumeGomez:
Then do, I'll merge it later !

GuillaumeGomez:
Crap, I thought we asked about import `std` instead of `std::ptr`. Let's say we didn't see anything.

GuillaumeGomez:
@gkoz: On this one, I don't agree with you. I think `to_glib` and `from_glib` are too general.

GuillaumeGomez:
@gkoz: Actually no, that could be nice.

PR #24, opened 2015-05-09
Object reform

gkoz:


gkoz:
To checkout all at once, as suggested in the RFC PR, you can
`repo init -u https://github.com/gkoz/rust-gnome-manifest -b object_reform`

PR #33, opened 2015-05-11
Add Pixbuf::new_from_vec that takes a Vec<u8> to use as the image data

gkoz:


gkoz:
Moved the `vec` argument to the first position where it belongs.

GuillaumeGomez:
Travis is ahving issue I think... Appveyor is good so I am too.

PR #45, opened 2015-05-20
GDK Cairo Interaction

### LINKED ISSUE #20: GDK Cairo Interaction
GDK offers a number of [methods for interfacing with Cairo surfaces](https://developer.gnome.org/gdk3/stable/gdk3-Cairo-Interaction.html).  I'd like to see these added, because they can be useful for loading images to render to Cairo surfaces (i.e. `gdk_cairo_set_source_pixbuf()`).  Where would be the best place to add these?  Should they be a part of **rust-gnome/gdk** or would they be better in **rust-gnome/cairo**?

jashephe:
This should largely resolve #20.  I can't say that I've tested it thoroughly, but it compiles, and I did a few spot checks.


GuillaumeGomez:
Nice work, thanks ! I think you forgot to reexport `GdkCairoInteraction` trait.

gkoz:
`GdkCairoInteraction` seems too long for a name that doesn't say anything. Some Rust conventions might suggest naming this `ContextExt` and the module could be `cairo` or `cairo_context`. And yes the trait should probably be reexported in the `prelude`.

gdk-sys/src/lib.rs
```diff
@@ -14,6 +14,7 @@ use libc::{c_int, c_char, c_double, c_void, c_uint, c_uchar, c_ulong, c_float};
 pub use cairo_ffi::cairo_rectangle_int_t as GdkRectangle;
 
 pub use glib_ffi::{gboolean, gpointer, gsize, GType};
+pub use cairo_ffi::{cairo_t, cairo_content_t, cairo_pattern_t, cairo_region_t, cairo_surface_t};
```
gkoz: I'm not sure reexporting a lot of types is needed. Even exporting `GdkRectangle` above might not have been a good idea.

src/cairo_interaction.rs
```diff
@@ -0,0 +1,83 @@
...
+//--> WRAP: gdk_cairo_region_create_from_surface (cairo_surface_t *surface);
+
+//pub fn create_surface_from_pixbuf() { }
+//--> WRAP: gdk_cairo_surface_create_from_pixbuf (const GdkPixbuf *pixbuf, int scale, GdkWindow *for_window);
+
+pub trait GdkCairoInteraction {
+    /// This is a convenience function around `clip_extents()`. It rounds
+    /// the clip extents to integer coordinates and returns a `GdkRectangle`,
+    /// or `None` if no clip area exists.
+    fn get_clip_rectangle(&self) -> Option<GdkRectangle>;
```
gkoz: I've been putting `cairo::RectangleInt` into native APIs meaning it to be the "canonical" name (although rustdoc recognizes no such thing currently) so if you don't mind that, we should be consistent.

src/cairo_interaction.rs
```diff
@@ -0,0 +1,83 @@
...
+    
+    //fn add_region(&self, region: ???);
+    //--> WRAP: fn gdk_cairo_region(cr: *mut cairo_t, region: *const cairo_region_t);
+}
+
+impl GdkCairoInteraction for Context {
+    fn get_clip_rectangle(&self) -> Option<GdkRectangle> {
+        let rectangle = ptr::null_mut();
+        unsafe {
+            match ffi::gdk_cairo_get_clip_rectangle(self.get_ptr(), rectangle) {
```
gkoz: I don't believe this can work. Passing null there just ignores that param. For an outparam you want 
```Rust
let mut rectangle = mem::uninitialized();
ffi::gdk_cairo_get_clip_rectangle(self.get_ptr(), &mut rectangle)
```

The `bool` specifics better be handled by `from_glib` making this more like
```Rust
if from_glib(ffi::gdk_cairo_get_clip_rectangle(self.get_ptr(), &mut rectangle)) {
    Some(rectangle)
}
else {
  None
}
```

src/cairo_interaction.rs
```diff
@@ -0,0 +1,83 @@
...
+            }
+        }
+    }
+
+    fn set_source_rgba(&self, rgba: &GdkRGBA) {
+        unsafe { ffi::gdk_cairo_set_source_rgba(self.get_ptr(), rgba); }
+    }
+
+    fn set_source_pixbuf(&self, pixbuf: &Pixbuf, x: f64, y: f64) {
+        unsafe { ffi::gdk_cairo_set_source_pixbuf(self.get_ptr(), pixbuf.to_glib_none().0, x as c_double, y as c_double); }
```
gkoz: I've long been wondering why people make no-op numeric casts explicit. `c_double` is `f64`, why the cast?

gdk-sys/src/lib.rs
jashephe: Okay, I was just trying to match what was done with __glib_ffi__, but I can take that out.

src/cairo_interaction.rs
jashephe: Sure, the majority of this was written before you changed the way `Rectangle`s are handled, so I figured that we might need a few changes after I merged in those changes.

src/cairo_interaction.rs
jashephe: Okay, I'll change it.

src/cairo_interaction.rs
jashephe: I saw it done in __rust-gnome/cairo__ and was just trying to use a consistent style.  I'll change it.

jashephe:
Okay, I can make these changes and re-export the trait.  I agree that `GdkCairoInteraction` is a poor name, but I didn't realize that we were willing to rename things from their GDK counterparts.  That might make it harder for some people to find the relevant C documentation.  Then again, we can probably just put a link to the official documentation in the comments or something.

GuillaumeGomez:
@jashephe: Yes, it would be preferable. In a more or less near future, we'll add all documentation comments (some parts already have it but a lot doesn't).

src/cairo_interaction.rs
gkoz: Yeah the style is inconsistent right now so it's understandable.

GuillaumeGomez: @gkoz: Didn't you harmonized the style in the previous days ? I must have miss something... I'm gonna take a look back.

gkoz:
> Okay, I can make these changes and re-export the trait. I agree that GdkCairoInteraction is a poor name, but I didn't realize that we were willing to rename things from their GDK counterparts.

We probably want to keep the names mostly but in this case I don't think it has any name in GDK.

src/cairo_interaction.rs
gkoz: Let's just change to the new type right away.

jashephe:
Okay, I've updated the files to reflect your feedback.  I renamed `GdkCairoInteraction` to `ContextExt` as suggested, but I elected to keep the module as `cairo_interaction`, since it also includes a few methods that shouldn't be implemented as traits on `Context`, and I felt that the name more closely correlates to the [relevant GDK3 documentation page](https://developer.gnome.org/gdk3/stable/gdk3-Cairo-Interaction.html).

GuillaumeGomez:
I don't have issue with the filename and the trait is reexported so everything's fine for me. @gkoz: do you see anything else ?

src/lib.rs
```diff
@@ -144,3 +145,7 @@ pub use gdk_ffi::enums::{
 pub use self::keys::{
     keyval_name
 };
+
+pub use cairo_interaction::{
+    ContextExt
```
gkoz: I think it's better to reexport in `prelude.rs` with the expectation that people will `use gdk::prelude::*;`

src/cairo_interaction.rs
```diff
@@ -0,0 +1,81 @@
...
+
+//! Functions to support using cairo
+
+use std::mem;
+use glib::translate::*;
+use ffi::{self, GdkRGBA};
+use super::{Pixbuf, Window};
+use cairo::{Context, RectangleInt};
+
+pub fn create(window: &Window) -> Context {
```
gkoz: What if we put this in the `ContextExt` trait as `create_from_window`?

jashephe: We could do that, but there are other functions on the [documentation page](https://developer.gnome.org/gdk3/stable/gdk3-Cairo-Interaction.html) like `create_region_from_surface()` and `create_surface_from_pixbuf()` which don't really fit in as extensions of `Context`.  I commented them out for now, because regions haven't been wrapped, but it's conceivable that they'll be added in the future.

gkoz: Since the module name isn't specific to `Context` those could go into similar traits for regions and surfaces it seems.

jashephe: That seems pretty reasonable to me.  I just changed it.

gkoz:
:+1:

GuillaumeGomez:
Perfect then, thanks for your work ! I merge.

PR #47, opened 2015-05-21
Added keyval_to_unicode

trolleyman:
Think I've got the hang of this now

### DIFF
--- a/gdk-sys/src/lib.rs
+++ b/gdk-sys/src/lib.rs
@@ -685,6 +685,7 @@ extern "C" {
     // Gdk Key Handling                                                  NOT OK
     //=========================================================================
     pub fn gdk_keyval_name                         (keyval:c_uint) -> *mut c_char;
+    pub fn gdk_keyval_to_unicode                   (keyval:c_uint) -> u32;
 
     //=========================================================================
     // GdkPixbufLoader                                                   NOT OK
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -7,6 +7,7 @@
 use glib::translate::*;
 use ffi;
 use libc::c_uint;
+use std::char;
 
 pub fn keyval_name(keyval: u32) -> Option<String> {
     unsafe {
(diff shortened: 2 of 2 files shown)


GuillaumeGomez:
Thanks for your PR ! You forgot the `use std::char;`. Could you also add an empty line between your declaration and your if / else please ? And lastly, could you squash your commits ? Normally I don't mind but 5 commits for 12 lines seems too much. I'll merge once it's done.

gdk-sys/src/lib.rs
```diff
@@ -685,6 +685,7 @@ extern "C" {
     // Gdk Key Handling                                                  NOT OK
     //=========================================================================
     pub fn gdk_keyval_name                         (keyval:c_uint) -> *mut c_char;
+    pub fn gdk_keyval_to_unicode                   (keyval:c_uint) -> u32;
```
gkoz: It would be consistent with the ffi style to use `c_uint` not `u32` here.

GuillaumeGomez: @gkoz: No, take a look [here](https://developer.gnome.org/gdk3/stable/gdk3-Keyboard-Handling.html#gdk-keyval-to-unicode). `u32` is perfectly fine.

gkoz: Yeah, okay, while `u32` is unusual I suppose it does reflect `guint32` more precisely.

GuillaumeGomez: When the size of the type is specified, I prefer to use its equivalent in Rust.

trolleyman:
Sure, one sec

GuillaumeGomez:
@trolleyman: Take your time, we're not in a hurry. ;-)

GuillaumeGomez:
Perfect like that ! (It misses the empty line but it's not important since the code is only 2 lines long). Can you squash your commits please ?

trolleyman:
Sorry but I have no idea how to do that, I've just been editing the files through the browser...

trolleyman:
I might just delete this PR and start again...?

trolleyman:
Yeah, I'll do that, I'll be back.

GuillaumeGomez:
Hum... Normally you just use this command :

```Shell
> git rebase -i HEAD~x
```

Where `x` is the number of commits which would have been squashed (in your case 6).

PR #48, opened 2015-05-21
Added keyval_to_unicode

trolleyman:
Maybe this will work this time! :P

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -16,8 +16,8 @@ pub fn keyval_name(keyval: u32) -> Option<String> {
 }
 
 pub fn keyval_to_unicode(keyval: u32) -> Option<char> {
-	unsafe {
-		let c = char::from_u32(ffi::gdk_keyval_to_unicode(keyval));
-		if c == Some('\0') { None } else { c }
-	}
+    unsafe {
+        let c = char::from_u32(ffi::gdk_keyval_to_unicode(keyval));
+        if c == Some('\0') { None } else { c }
+    }
 }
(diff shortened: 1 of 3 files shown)


src/keys.rs
```diff
@@ -7,9 +7,17 @@
...
 
 pub fn keyval_name(keyval: u32) -> Option<String> {
     unsafe {
         from_glib_none(ffi::gdk_keyval_name(keyval as c_uint))
     }
 }
+
+pub fn keyval_to_unicode(keyval: u32) -> Option<char> {
+	unsafe {
+		let c = char::from_u32(ffi::gdk_keyval_to_unicode(keyval));
```
GuillaumeGomez: We use spaces tabulation (' ') not tab tabulations ('\t'). On top of that, your indentation isn't good. Not good ! =D

src/keys.rs
```diff
@@ -7,9 +7,17 @@
...
+		if c == Some('\0') {None} else {c}
```
GuillaumeGomez: Put spaces insides brackets :

```Rust
if c == Some('\0') { None } else { c }
```

GuillaumeGomez:
Fix the issues I mentioned and please squash like I explained [here](https://github.com/rust-gnome/gdk/pull/47#issuecomment-104252508).

trolleyman:
Arg, thought I was on spaces :P

GuillaumeGomez:
I understand, I had the same issues when I started. ^^

PR #50, opened 2015-05-23
Adding functions for Pixbuf scaling

jashephe:
Added functions for scaling `Pixbuf`s as described in "[Scaling](https://developer.gnome.org/gdk-pixbuf/unstable/gdk-pixbuf-Scaling.html)".

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -140,7 +140,7 @@ pub use gdk_ffi::enums::{
     Gravity,
     DragAction,
     DragProtocol,
-    InterpType
+    InterpType,
 };
 
 pub use self::keys::{
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -139,7 +139,8 @@ pub use gdk_ffi::enums::{
     WindowWindowClass,
     Gravity,
     DragAction,
-    DragProtocol
+    DragProtocol,
+    InterpType
 };
 
 pub use self::keys::{
--- a/src/pixbuf/mod.rs
+++ b/src/pixbuf/mod.rs
@@ -247,10 +247,11 @@ impl Pixbuf {
     }
 
     /// Flips a pixbuf horizontally or vertically and returns the result in a
-    /// new pixbuf, or `None` if not enough memory could be allocated for it.
-    pub fn flip(&self, horizontal: bool) -> Option<Pixbuf> {
+    /// new pixbuf, or `Err` if not enough memory could be allocated for it.
+    pub fn flip(&self, horizontal: bool) -> Result<Pixbuf, ()> {
         unsafe {
-            from_glib_none(ffi::gdk_pixbuf_flip(self.to_glib_none().0, horizontal.to_glib()))
+            Option::from_glib_full((ffi::gdk_pixbuf_flip(self.to_glib_none().0,
+                                                         horizontal.to_glib()))).ok_or(())
         }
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -139,7 +139,8 @@ pub use gdk_ffi::enums::{
     WindowWindowClass,
     Gravity,
     DragAction,
-    DragProtocol
+    DragProtocol,
```
gkoz: This is why trailing commas are recommended so adding or removing an element doesn't touch other lines.

src/pixbuf/mod.rs
```diff
@@ -181,6 +183,66 @@ impl Pixbuf {
...
+    /// scaling down `Nearest` is usually unusably ugly). The default
+    /// `interp_type` should be `Bilinear` which offers reasonable quality and
+    /// speed.
+    ///
+    /// You can scale a sub-portion of `src` by creating a sub-pixbuf pointing
+    /// into src ; see `new_subpixbuf()`.
+    ///
+    /// For more complicated scaling/compositing see `scale()` and
+    /// `composite()`.
+    pub fn scale_simple(&self, dest_width: i32, dest_height: i32, interp_type: InterpType) -> Option<Pixbuf> {
```
gkoz: I believe we've settled on using `Result` for failures. At least that's what `new` does.
We could rethink this particular failure mode and just panic on the OOM failures.

src/pixbuf/mod.rs
```diff
@@ -181,6 +183,66 @@ impl Pixbuf {
...
+        unsafe {
+            from_glib_none(ffi::gdk_pixbuf_scale_simple(self.to_glib_none().0, dest_width, dest_height, interp_type))
```
gkoz: The sys crates are a lost cause but otherwise let's try to follow the 100-column rule.

GuillaumeGomez:
@gkoz: Do you want it to be merged like that ? The only real problem is the `Option` stuff but since we didn't start to change the stuff, does it matter for now ?

gkoz:
The Pixbuf part seems to have been converted to Results. I guess we could do it but pushing to crates like this would commit the API until the next push when we would break it once again.

GuillaumeGomez:
I can wait, we're not in a hurry.

src/pixbuf/mod.rs
```diff
@@ -181,6 +183,77 @@ impl Pixbuf {
...
+        unsafe {
+            ffi::gdk_pixbuf_composite(self.to_glib_none().0, dest.to_glib_none().0, dest_x, dest_y,
+                                      dest_width, dest_height, offset_x, offset_y, scale_x,
+                                      scale_y, interp_type, overall_alpha);
+        }
+    }
+
+    /// Flips a pixbuf horizontally or vertically and returns the result in a
+    /// new pixbuf, or `None` if not enough memory could be allocated for it.
+    pub fn flip(&self, horizontal: bool) -> Option<Pixbuf> {
```
gkoz: This one seems to be missing the same `Result` treatment.

jashephe: Indeed.  Give me a sec.

gkoz:
Looks good!

GuillaumeGomez:
Perfect, I merge. Thanks for your work ! =D

PR #57, opened 2015-07-01
Documentation

GuillaumeGomez:


src/atom.rs
```diff
@@ -2,6 +2,8 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
+//! Atoms — Functions to manipulate properties on windows
```
gkoz: We probably won't publish the modules, only reexport types from them so this kind of annotation seems better suited to the structs not modules.

GuillaumeGomez: I was thinking about. I'll move the doc to the type.

GuillaumeGomez:
cc @gkoz

src/device_manager.rs
```diff
@@ -11,6 +9,103 @@ use display::Display;
...
+/// Virtual devices are always paired, so there is a keyboard device for every pointer device.
+/// Associations between devices may be inspected through Device::get_associated_device().
+///
+/// There may be several virtual devices, and several physical devices could be controlling each of
+/// these virtual devices. Physical devices may also be “floating”, which means they are not attached
+/// to any virtual device.
+///
+/// #Master and slave devices
+///
+/// ``ìgnore
```
gkoz: Does this part render correctly? The syntax looks weird: two backticks never terminated?

src/device.rs
```diff
@@ -22,68 +21,158 @@ impl types::StaticType for Device {
...
+    ///     gchar *path;
+    /// 
+    ///     let vendor = device.get_vendor_id().unwrap();
+    ///     let product = device.get_product_id().unwrap();
+    /// 
+    ///     let path = format!("/org/example/app/devices/{}:{}/", vendor, product);
+    ///     g_settings_new_with_path(DEVICE_SCHEMA, path);
+    /// }
+    /// ```
+    #[cfg(feature = "gdk_3_16")]
```
gkoz: Looks like these need a `s/feature = "([^"]+)"/$1/`.

src/device_manager.rs
GuillaumeGomez: I see what happened, it became another character.

src/device.rs
GuillaumeGomez: What ? I don't understand your comment.

gkoz: The current syntax is `cfg(gdk_3_16)` and that regex is how I'd do the conversion.

GuillaumeGomez: Oh right. I forgot we changed the system !

GuillaumeGomez:
Updated.

gkoz:
Looks good!

src/device_manager.rs
```diff
@@ -11,6 +9,103 @@ use display::Display;
...
+///
+/// There may be several virtual devices, and several physical devices could be controlling each of
+/// these virtual devices. Physical devices may also be “floating”, which means they are not attached
+/// to any virtual device.
+///
+/// #Master and slave devices
+///
+/// ```ignore
+/// carlos@sacarino:~$ xinput list
+/// ⎡ Virtual core pointer                          id=2    [master pointer  (3)]
```
gkoz: `rustdoc` chokes on this :(

PR #60, opened 2015-07-19
Replace features with cfg and automatic version detection

gkoz:
The user is no longer required to set the target library version, the builds script detects it from `pkg-config` output and sets the appropriate `cfg` directives.

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -236,7 +236,6 @@ impl Display {
     /// Currently this only works on X11 with XComposite and XDamage extensions available.
     ///
     /// Deprecated since GDK 3.16
-    #[cfg(not(gdk_3_16))]
     pub fn supports_composite(&self) -> bool {
         unsafe { from_glib(ffi::gdk_display_supports_composite(self.to_glib_none().0)) }
     }
--- a/src/display.rs
+++ b/src/display.rs
@@ -236,7 +236,7 @@ impl Display {
     /// Currently this only works on X11 with XComposite and XDamage extensions available.
     ///
     /// Deprecated since GDK 3.16
-    #[cfg(not(feature = "gdk_3_16"))]
+    #[cfg(not(gdk_3_16))]
     pub fn supports_composite(&self) -> bool {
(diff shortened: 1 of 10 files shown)


gkoz:
cc @EPashkin

gkoz:
Similar to https://github.com/rust-gnome/gtk/pull/120 but not muddied with the `gtk_glue.c` handling, which we won't need going forward.

src/display.rs
```diff
@@ -236,7 +236,6 @@ impl Display {
     /// Currently this only works on X11 with XComposite and XDamage extensions available.
     ///
     /// Deprecated since GDK 3.16
-    #[cfg(not(feature = "gdk_3_16"))]
```
GuillaumeGomez: How do you handle this then ? I didn't see anything who did remove this function from the build if `gdk` version was >= 3.16.

gkoz: When the features were set manually, this attribute made some sense, though I'd still disagree with removing functions that have been deprecated so recently. With automatic detection it would set the users up for surprising breakage caused by simply updating their libraries -- a big no-no.

GuillaumeGomez: The problem here is that this function doesn't exist anymore from 3.16. Not just deprecated.

gkoz: They did what!?

GuillaumeGomez: Oh, my bad. Only deprecated in 3.16.

gkoz: [It's there all right](https://git.gnome.org/browse/gtk+/tree/gdk/gdkdisplay.h#n163).

GuillaumeGomez: Yes, that's what I said. :-) Then everything's good. I merge (and update crates ?).

gkoz: The comments don't always get loaded automatically... :)

GuillaumeGomez: This is dark magic ! ;)

gkoz:
Let's not update yet, let this sit in master for a week. Also cleaning up features is not done with this.

GuillaumeGomez:
Ouf ! It took me an eternity yesterday ! T_T

PR #62, opened 2015-07-29
Use the glib::translate API with cairo::Context

gkoz:


GuillaumeGomez:
Thanks for this update ! I wait appveyor status and then I merge !

PR #66, opened 2015-08-13
Fix the build, followup to 4fc94268

gkoz:
This time let's wait for the green.


gkoz:
So there still is breakage in `gtk` but this commit seems to fulfil its purpose.

PR #69, opened 2015-09-06
Mark the non-rust 'code' snippet as text

gkoz:


GuillaumeGomez:
Thanks for the update !

PR #71, opened 2015-09-08
Remove duplicated method

Flakebi:
Hello,
compiling gdk (and gtk and so on) failed for me because `get_vendor_id` is defined twice. Maybe the second method is at the wrong place (the other method is referenced in the documentation comment)?

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -28,7 +28,7 @@ impl Device {
         }
     }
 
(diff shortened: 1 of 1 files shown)


gkoz:
Thanks for reporting. It looks like the second method was supposed to be `get_product_id`, so it needs to be renamed not removed.

GuillaumeGomez:
See #73

Flakebi:
Thanks for fixing the issue!

PR #75, opened 2015-10-03
Add GLContext widget

GuillaumeGomez:


gkoz:
It's a shame you had to do this manually cuz there's nothing here the generator couldn't do (remember `gdk` is "reformed" already).

src/gl_context.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    /// Retrieves the GLContext that this context share data with.
+    pub fn get_shared_context(&self) -> Option<Window> {
+        unsafe { from_glib_none(ffi::gdk_gl_context_get_shared_context(self.to_glib_none().0)) }
+    }
+
+    /// Retrieves the OpenGL version of the context.
+    ///
+    /// The context must be realized prior to calling this function.
+    pub fn get_version(&self, major: &mut i32, minor: &mut i32) {
```
gkoz: The generator will return these in a tuple so to avoid breakage I'd suggest doing the same.
Example: https://github.com/gkoz/gir/pull/122#issuecomment-143478279

src/lib.rs
```diff
@@ -85,6 +85,8 @@ pub use pixbuf::loader::PixbufLoader;
 pub use screen::Screen;
 pub use visual::Visual;
 pub use window::Window;
+#[cfg(gdk_3_16)]
+pub use gl_context::GLContext;
```
gkoz: Missing `mod gl_context`?

GuillaumeGomez:
It does ?! Impressive !

gkoz:
Well, I have to admit it doesn't add the docs yet.

GuillaumeGomez:
Too bad. Well, I think we'll still need to add it ourselves. (Updated btw)

src/gl_context.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub fn get_shared_context(&self) -> Option<Window> {
+        unsafe { from_glib_none(ffi::gdk_gl_context_get_shared_context(self.to_glib_none().0)) }
+    }
+
+    /// Retrieves the OpenGL version of the context.
+    ///
+    /// The context must be realized prior to calling this function.
+    pub fn get_version(&self) -> (i32, i32) {
+        let major = 0;
+        let minor = 0;
```
gkoz: `mut`?

GuillaumeGomez:
That's why I hate writing without having the possibility to test it...

gkoz:
Looks like we'll have to learn to build this stuff from source anyway. And a chroot is not too difficult to set up.

gkoz:
It's not an `Object`, till you've implemented `StaticType` for it [like this](https://github.com/gtk-rs/gdk/blob/master/src/window.rs#L112-L114).

I know, this boilerplate begs for a macro. I was hoping to avoid them previously. Looks like macros `impl_traitWidget` style are winning after all. But for now that's how it is.

src/gl_context.rs
```diff
@@ -0,0 +1,153 @@
...
+        unsafe { from_glib_none(ffi::gdk_gl_context_get_display(self.to_glib_none().0)) }
+    }
+
+    /// Retrieves the GdkWindow used by self.
+    pub fn get_window(&self) -> Option<Window> {
+        unsafe { from_glib_none(ffi::gdk_gl_context_get_window(self.to_glib_none().0)) }
+    }
+
+    /// Retrieves the GLContext that this context share data with.
+    pub fn get_shared_context(&self) -> Option<Window> {
```
gkoz: This returns `Option<GLContext>` perhaps?

src/gl_context.rs
```diff
@@ -0,0 +1,153 @@
+// Copyright 2013-2015, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+//! OpenGL context
+
+use glib::translate::*;
+use glib::types;
+use cursor::Cursor;
```
gkoz: unused import

src/gl_context.rs
```diff
@@ -0,0 +1,153 @@
...
+use display::Display;
+use object::Object;
+use screen::Screen;
```
gkoz: unused import

gkoz:
This is good at last! :)

PR #82, opened 2015-11-16
Assert that GDK is initialized in constructors and free functions

gkoz:
The first stab at fixing https://github.com/gtk-rs/gtk/issues/188


src/rt.rs
```diff
@@ -44,49 +97,60 @@ pub fn set_allowed_backends(backends: &str) {
...
 
 pub fn set_program_class(program_class: &str) {
+    assert_initialized_main_thread!();
     unsafe {
         ffi::gdk_set_program_class(program_class.to_glib_none().0)
     }
 }
 
 pub fn flush() {
+    assert_initialized_main_thread!();
```
GuillaumeGomez: This is exactly what I was afraid of. We now have to call the macro everywhere. :-/

GuillaumeGomez:
Thanks for this. I'm really not comfortable here since we might forget places from where it could need it... We already told about this, so let's say this is a temporary solution while we try to figure out something better.

gkoz:
I'll look into making an automated script, that would keep us from forgetting to put the assert everywhere it's needed. Adding this to `gtk` by hand is a nonstarter anyway. ;)

GuillaumeGomez:
I absolutely agree with you, it'd take an eternity or two to do it. Can't we add it to one of the macros which generates structs (widgets and others) ? Or do you want an external script ?

gkoz:
I'm afraid there's just no centralized place to put this. The generator will need to be adjusted to emit the assert of course but this won't cover everything. I'll see if my favorite tool, PCRE, is going to be helpful here.

gkoz:
Anyway don't merge this yet, I forgot to reexport `set_initialized`.

GuillaumeGomez:
Just ping me when it's done then. :)

gkoz:
Added a script to check any safe non-method fn has an assertion or skips one explicitly. This has actually caught several functions I've missed previously so you were absolutely right to be worried.

src/cairo_interaction.rs
```diff
@@ -10,10 +10,10 @@ use ffi::{self, GdkRGBA};
 use super::{Pixbuf, Window};
 use cairo::{Context, RectangleInt};
 
-//pub fn create_region_from_surface() { }
+//pub fn create_region_from_surface()
```
GuillaumeGomez: Your script deletes "{ }" too ? Is it a vonlontary behavior ?

GuillaumeGomez:
Also, what happens if the user quits the `gtk_main()` function and then relaunch it ? I'm not sure of what will happen in a C code, I should test later.

src/cairo_interaction.rs
gkoz: I had to delete those to avoid false positives.

gkoz:
`gtk_main` doesn't perform any destruction or significant initialization (it it unsafe to call `gtk_main_quit` outside the main loop so I'm going to add a check there).

src/cairo_interaction.rs
gkoz: The included script doesn't attempt to modify the files it only prints any offending functions.

GuillaumeGomez:
Always wonderful little annoying tricks to handle... Poor us :'(

src/cairo_interaction.rs
GuillaumeGomez: Oh okay, I thought it did both. I didn't look enough.

gkoz:
@GuillaumeGomez okay, this one seems ready, it has to go in before the gtk counterpart.

GuillaumeGomez:
Okay, I merge. Thanks !

PR #83, opened 2015-11-19
Reexport GdkRectangle

EPashkin:


GuillaumeGomez:
Thanks !
cc @gkoz

gkoz:
The background: `GdkRectangle` is an alias of `cairo_rectangle_int_t`, which is reexported in `cairo` as `RectangleInt`. This crate reexports `RectangleInt` and uses it in the APIs in place of `GdkRectangle`.

That decision was made before the move to automation so it might not make as much sense now. We could reexport it as `Rectangle`. Which crate we reexport it from might not even make a difference, rustdoc will probably still link to `cairo_rectangle_int_t`.

EPashkin:
@GuillaumeGomez seems direct reexport not good solution, do not merge this.

GuillaumeGomez:
It seemed weird to me and I wanted @gkoz's opinion. So I close it.

PR #105, opened 2016-03-02
Export GDK_KEY constants as u32.

rillian:
Key values are generally guint or guint32 in API calls,
but gir defines the GDK_KEY_* contants as gint, forcing
every use to be cast. This is probably just the generator
assuming gint for #define contants, and the signed vs.
unsigned distinction doesn't matter in C or most scripting
languages. But it does matter in rust.

Special-casing these in the the gir generator, or marking
up the gdk C header to add a U literal specifier everywhere
is the ideal fix, but in the meantime, we can hide the
bug by casting when we export the ffi version of the constants.


gkoz:
Good analysis! Can we take this further though? Is there value in keeping the type `u32` rather than making an enum?

rillian:
An enum would be more natural for rust, but I don't know how to handle keysym aliasing. That is, about 2000 GDK_KEY contants reference the same value as another constant. Rust enums don't allow this. E.g.
```
/usr/include/gtk-3.0/gdk/gdkkeysyms.h:#define GDK_KEY_Armenian_question 0x100055e
/usr/include/gtk-3.0/gdk/gdkkeysyms.h:#define GDK_KEY_Armenian_paruyk 0x100055e
```

gkoz:
Okay, looks like aliasing and non-exhaustiveness (there're likely to be values in the wild without corresponding constants) make an enum unsuitable.

Let's merge this then. @GuillaumeGomez

GuillaumeGomez:
@gkoz: I did an enum in the old API but I had then to create an interface function to convert from rust to C, so the current situation is better I think (a bit annoying but we can't do much about it...).

Thanks @rillian!

PR #107, opened 2016-03-02
Add getters for Event properties.

hmeyer:


gkoz:
Thanks for the PR! I'm adding an issue #108 with a higher level overview of the goals and some more specific comments here:

Since `window` is a member of `EventAny` it should be possible to add just one accessor to `Event`, all the others will get the method via autoderef.

`EventKey` documentation seems to strongly discourage using `string`.

hmeyer:
I addressed most comments.
In EventSetting I still use std::ffi::CStr - shouldn't that be safe?

gkoz:
`CStr` is safe but puts a burden of converting to `&str` on the user. So far we have been choosing to handle that in the library.

Take `GdkEventSetting::name`: <strike>it's probably reasonable to expect it to be ASCII</strike> ...actually I can't make any argument about this event. I don't see any references to in the [GIRs](https://github.com/gtk-rs/gir-files). We might do well to just delete it.

src/event_key.rs
```diff
@@ -2,10 +2,42 @@
...
+        self.as_ref().keyval
+    }
+
+    pub fn get_length(&self) -> u32 {
+        let length = self.as_ref().length;
+        assert!(length >= 0, "Unexpected negative value");
+        length as u32
+    }
+
+    pub fn get_hardware_keycode(&self) -> ::enums::key::Key {
```
gkoz: I think this and `get_keyval` need to be the other way around: `enums::key` is about keyvals and `Key` is `u32`, while `hardware_keycode` is `u16`.

src/event_configure.rs
```diff
@@ -9,3 +9,19 @@ pub struct EventConfigure(::Event);
 
 event_wrapper!(EventConfigure, GdkEventConfigure);
 event_subtype!(EventConfigure, Configure);
+
+impl EventConfigure {
+    pub fn get_position(&self) -> (u32, u32) {
```
gkoz: Coordinates probably can legitimately have negative values.

src/event_owner_change.rs
```diff
@@ -9,3 +9,25 @@ pub struct EventOwnerChange(::Event);
...
+impl EventOwnerChange {
+    pub fn get_owner(&self) -> Option<::Window> {
+        unsafe { from_glib_none(self.as_ref().owner) }
+    }
+
+    pub fn get_reason(&self) -> ::OwnerChange {
+        self.as_ref().reason
+    }
+
+    pub fn get_selection(&self) -> Option<::Atom> {
```
gkoz: `Atom` is tricky. It has a `NONE` value so it doesn't need `Option`.

src/event_grab_broken.rs
```diff
@@ -9,3 +9,17 @@ pub struct EventGrabBroken(::Event);
 
 event_wrapper!(EventGrabBroken, GdkEventGrabBroken);
 event_subtype!(EventGrabBroken, GrabBroken);
+
+impl EventGrabBroken {
+    pub fn is_keyboard(&self) -> bool {
+        self.as_ref().keyboard != 0
```
gkoz: Oh, and I'd prefer bool conversions to be consistent with the rest of the code:
```
from_glib(self.as_ref().keyboard)
```

hmeyer:
next round...

src/event_window_state.rs
```diff
@@ -9,3 +9,13 @@ pub struct EventWindowState(::Event);
 
 event_wrapper!(EventWindowState, GdkEventWindowState);
 event_subtype!(EventWindowState, WindowState);
+
+impl EventWindowState {
+    pub fn get_changed_mask(&self) -> ::WindowState {
+        self.as_ref().changed_mask
+    }
+
+    pub fn get_new_state(&self) -> ::WindowState {
```
gkoz: Naming inconsistency here doesn't seem justified.

gkoz:
There are a couple of redundant `extern crate` lines. We already have those crates imported in `lib.rs`, so it's enough to `use` them or specific types (e.g. `use cairo::RectangleInt;`).

Other than these two comments seems good to go.

hmeyer:
done.

gkoz:
Thanks for contributing @hmeyer!

PR #115, opened 2016-04-18
Enums

EPashkin:
Part of https://github.com/gtk-rs/gtk/issues/307


EPashkin:
@gkoz Please, look at changed functions with `uninitialized`
IMHO PR ready for merge.

src/device.rs
```diff
@@ -53,11 +54,15 @@ impl Device {
...
-        unsafe { ffi::gdk_device_set_key(self.to_glib_none().0, index_, keyval, modifiers) }
+        unsafe { ffi::gdk_device_set_key(self.to_glib_none().0, index_, keyval, modifiers.to_glib()) }
     }
 
-    pub fn get_key(&self, index_: u32, keyval: &mut u32, modifiers: &mut ::ModifierType) -> bool {
-        unsafe { from_glib(ffi::gdk_device_get_key(self.to_glib_none().0, index_, keyval, modifiers)) }
+    pub fn get_key(&self, index_: u32, keyval: &mut u32) -> (bool,::ModifierType) {
+        unsafe {
+          let mut modifiers = mem::uninitialized();
+          let ret = from_glib(ffi::gdk_device_get_key(self.to_glib_none().0, index_, keyval, &mut modifiers));
```
gkoz: This looks like a typical case of out-params + boolean, returning `Option<(u32, ModifierType)>`.

gkoz:
We'll also want `Gdk.GLError`.

This part seems easy and harmless enough. Are the remaining reexports in `lib.rs` apparently unused? There seem to be too many of them.

Gir.toml
```diff
@@ -1,11 +1,19 @@
 [options]
-girs_dir = "gir-files"
+girs_dir = "../gir-files"
 library = "Gdk"
 version = "3.0"
 min_cfg_version = "3.4"
 target_path = "."
 work_mode = "normal"
-generate_safety_asserts = true
+#generate_safety_asserts = true
```
gkoz: We have the same set of asserts here as in `gtk`.

EPashkin:
Fixed, also updated `gdk_window_get_device_position` and `gdk_window_get_device_position_double` as seems returning `Window` and returning other independent.

Gir.toml
```diff
@@ -1,12 +1,21 @@
 [options]
-girs_dir = "gir-files"
+girs_dir = "../gir-files"
```
gkoz: Since we're generating some code now it makes sense to add the usual submodules and makefile.

src/window.rs
```diff
@@ -493,24 +493,29 @@ impl Window {
...
-                device.to_glib_none().0,
-                x, y, mask))
+            let mut x = mem::uninitialized();
+            let mut y = mem::uninitialized();
+            let mut mask = mem::uninitialized();
+            let ret = from_glib_none(
+                ffi::gdk_window_get_device_position_double(self.to_glib_none().0,
+                                                           device.to_glib_none().0,
+                                                           &mut x, &mut y, &mut mask));
+            (ret, x, y, from_glib(mask))
```
gkoz: Yep, this seems to be more correct than before. I wonder if the generator would do the same...

EPashkin: Its copy-pasted from generated code ;)

EPashkin:
Added submodules.

EPashkin:
I strongly don't like idea of having many `gir` sub-modules ,
as it slow to rebuild and contains "big" `target`,
so IMHO we need rethink sub-module structure.

EPashkin:
Updated makefile

EPashkin:
Ready for merge

GuillaumeGomez:
Good for me.

@gkoz: Merge if you think it's all good.

EPashkin:
Added `CursorType`

GuillaumeGomez:
This PR is getting bigger and bigger. O.O

EPashkin:
Yes, it too many unneeded additions

PR #118, opened 2016-05-13
Publish EventAny under its name

EPashkin:
To generate `Widget` signals `map-event` and `unmap-event` in https://github.com/gtk-rs/gtk/pull/339

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -100,6 +100,7 @@ pub use display::Display;
 pub use display_manager::DisplayManager;
 pub use drag_context::DragContext;
 pub use event::Event;
+pub use event::Event as EventAny;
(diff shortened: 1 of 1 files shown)


gkoz:
A hack like this just to make two signals automatic feels dubious.

EPashkin:
Then what we do with 2 problem signals now?

gkoz:
They can live in the manual widget trait, it's there anyway.

EPashkin:
So only 2 signals in `WidgetSignals` :(

gkoz:
I meant `WidgetExtManual` because I assume we're getting rid of all `*Signal` traits.

EPashkin:
Ok that better that restoring it in `signals.rs`

PR #127, opened 2016-07-23
WindowAttr

### LINKED ISSUE #126: Fix struct WindowAttr visibility
It's not exported in `lib.rs`.

EPashkin:
Rename `Attributes` to `WindowAttr`, implement `Default`
Fixes #126
Related to https://github.com/gtk-rs/gtk/issues/366

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -42,7 +42,26 @@ pub struct WindowAttr {
     pub type_hint: Option<WindowTypeHint>,
 }
 
-impl Attributes {
+impl Default for WindowAttr {
+    fn default() -> WindowAttr {
+        skip_assert_initialized!();
+        WindowAttr {
+            title: None,
+            event_mask: 0,
+            x: None,
+            y: None,
+            width: 400,
+            height: 300,
+            wclass: WindowWindowClass::InputOutput,
+            visual: None,
+            window_type: WindowType::Toplevel,
+            cursor: None,
+            override_redirect: false,
+            type_hint: None,
+        }
+    }
+}
+
 impl WindowAttr {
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
:+1:

src/window.rs
```diff
@@ -42,7 +42,26 @@ pub struct Attributes {
...
-impl Attributes {
+impl Default for WindowAttr {
+    fn default () -> WindowAttr {
+        WindowAttr {
+            title: None,
+            event_mask: 0,
+            x: None,
+            y: None,
+            width: 800,
+            height: 600,
```
gkoz: These a are bit concerning to me. Wouldn't 'neutral' values be smaller somewhere between 100 and 400?

EPashkin: 400x300 ?

gkoz: :+1: 

BTW, this might be a good occasion to make these fields unsigned.

EPashkin: And `event_mask` ?

EPashkin: X,Y?

EPashkin: No, IMHO x and y can be negative

EPashkin: After checking, too many width\heights signed so better don't make unsigned in one place.

gkoz:
Apparently `check_init_asserts` is unhappy.

EPashkin:
Updated

GuillaumeGomez:
Seems good for me. I'll let @gkoz handles the merge. ;)

gkoz:
Thanks!

PR #130, opened 2016-11-05
Apply last gir

EPashkin:
Changes from https://github.com/gtk-rs/gir/pull/251 and https://github.com/gtk-rs/gir/pull/243


GuillaumeGomez:
Thanks!

PR #139, opened 2017-03-04
First draft implementation of RGBA

Susurrus:
I want to implement `Gdk::RGBA` as this is a pretty big missing type needed in GTK. I tried adding it to the auto-generated list, but I get the following error:
```
73 |             from_glib(ffi::gdk_rgba_equal(self.to_glib_none().0, p2.to_glib_none().0))
   |                                           ^^^^^^^^^^^^^^^^^^^^^ expected enum `libc::c_void`, found struct `ffi::GdkRGBA`
```

So I started modifying the existing RGBA implementation. And manually cast these arguments are `gconstpointer` so I think I just use `as gconstpointer` here, which is what I did, please correct me if that's not right. Is this a change that should be done in `gir` instead?

Also, can someone confirm that `GdkRGBA` should be `Boxed` type? I think it should since the GDK library implements `copy` and `free`, but the docs weren't very clear on how to choose the appropriate type for `glib_wrapper!`.

Additionally, the color functions that exist (`white`, `blue`, etc.) don't exist in `GDK` and I don't know how to initialize them given the wrapped type. Can someone a) confirm that we want to keep these color functions and if so b) explain how to initialize them?


EPashkin:
`gconstpointer` is good.
`white`, ... seems better do with `parse`

GuillaumeGomez:
@EPashkin: I like the current white method.

EPashkin:
If do it with parse then we can use constructor
```diff
@@ -67,6 +69,16 @@ impl RGBA {
         }
     }
 
+fn from_spec(spec: &str) -> Option<RGBA> {
+unsafe {
+    let mut ffiRGBA = mem::uninitialized();
+    if from_glib(ffi::gdk_rgba_parse(&mut ffiRGBA as *mut _, spec.to_glib_none().0)) {
+        Some(from_glib_full(ffi::gdk_rgba_copy(&ffiRGBA)))
+    } else {
+        None
+    }
+}
+}
     fn equal(&self, p2: &RGBA) -> bool {
         unsafe {
             from_glib(ffi::gdk_rgba_equal(self.to_glib_none().0 as gconstpointer,
@@ -88,7 +100,7 @@ impl RGBA {
 
     fn to_string(&self) -> String {
         unsafe {
-            from_glib_full(ffi::gdk_rgba_to_string(self.to_glib_none_mut().0))
+            from_glib_full(ffi::gdk_rgba_to_string(self.to_glib_none().0))
         }
     }
 }
```
If don't use parse then as `RGBA` has `free` then we needed other constructor that create ffiRGBA from parts and copy it and convert

EPashkin:
IMHO we really needed both constructors

EPashkin:
@GuillaumeGomez I like `white` method too, but better realize it with calling other constructor.

EPashkin:
Also IMHO better generate `Gdk.RGBA` so there will be two files `src/auto/r_g_b_a.rs` and `src/rgba.rs` with manual parts: 2 constructors, white, ...
And all constructors must have `skip_assert_initialized!();`

GuillaumeGomez:
Hum, I see your point. I agree! :)

Susurrus:
Okay, I don't quite follow everything that was said here, but I'll try and enumerate everything here:

1. We'll use `gir` to autogenerate RGBA. Right now this doesn't work because of the `gconstpointer` cast. Looks like this is a change required in `gir`, is that correct?
2. We want two constructors. Is one a `new(red, green, blue, alpha)` and the other is the `from_spec()` method @EPashkin already posted?
3. We will keep the `white()` and related functions. These will use the `new(r, g, b, a)` constructor to create an return a new `Gdk::RGBA`.
4. I'd still like confirmation that his is a `Boxed` type. I wasn't quite clear if it was.

EPashkin:
It really `Boxed` type as it has `free` and `copy`
2 and 3 as you wrote.
You right about `gconstpointer` I missed that,
I think about changes in gir,
currently you can don't generate these 2 functions and implement its and `Eq` manually
```
[[object]]
name = "Gdk.RGBA"
status = "generate"
    [[object.function]]
    name = "equal"
    ignore = true
    [[object.function]]
    name = "hash"
    ignore = true
```
Also forgetting to remove `pub use ffi::GdkRGBA as RGBA;` in `lib.rs` can cause strange error as compiler prefer it but no `auto::r_g_b_a::RGBA`

These changes needed to `gtk` to builds, but actually `get/set_rgba` must go to `src\color_chooser.rs` that must work on `gdk::RGBA` and not on `gdk_ffi::GdkRGBA`.
Currently adding `Gdk.RGBA` as manual  in `gtk` causes many errors and needed or fixing `glib::Value::from` or ignoring properties and IMHO not worth on current stage.
```diff
--- a/src/color_button.rs
+++ b/src/color_button.rs
@@ -21,7 +21,7 @@ impl ColorButton {
     pub fn new_with_rgba(rgba: &gdk::RGBA) -> ColorButton {
         assert_initialized_main_thread!();
         unsafe {
-            Widget::from_glib_none(ffi::gtk_color_button_new_with_rgba(rgba)).downcast_unchecked()
+            Widget::from_glib_none(ffi::gtk_color_button_new_with_rgba(rgba.to_glib_none().0)).downcast_unchecked()
         }
     }
 
@@ -35,8 +35,8 @@ impl ColorButton {
 
     pub fn get_rgba(&self) -> gdk::RGBA {
         unsafe { 
-            let mut rgba = mem::uninitialized();
-            ffi::gtk_color_button_get_rgba(self.to_glib_none().0, &mut rgba);
+            let mut rgba = gdk::RGBA::uninitialized();
+            ffi::gtk_color_button_get_rgba(self.to_glib_none().0, rgba.to_glib_none_mut().0);
             rgba
         }
     }
@@ -46,6 +46,6 @@ impl ColorButton {
     }
 
     pub fn set_rgba(&self, rgba: &gdk::RGBA) {
-        unsafe { ffi::gtk_color_button_set_rgba(self.to_glib_none().0, rgba) }
+        unsafe { ffi::gtk_color_button_set_rgba(self.to_glib_none().0, rgba.to_glib_none().0) }
     }
 }
```

Susurrus:
Here's another implementation that is almost ready, it just needs the `Gdk::RGBA::new()` part finished. I couldn't figure out how to instantiate a new struct while using the `glib::Boxed` type, so right now compilation fails. Can someone clarify how I should actually be generating a new `ffi::GdkRGBA` object and keep it under GLib's memory management? That should be the last missing piece for the Gdk work. 

I'm aware there's more work necessary for GTK and other libraries, which I'll work on once I finished this PR so I can easily test locally.

src/rgba.rs
```diff
@@ -2,90 +2,76 @@
...
-            red: 1f64,
-            green: 1f64,
-            blue: 1f64,
-            alpha: 1f64
+        ffi::GdkRGBA {
+            red: red,
+            green: green,
+            blue: blue,
+            alpha: alpha,
         }
```
EPashkin: Forgot `gdk_rgba_copy`, this version causes memory error on drop

src/rgba.rs
```diff
@@ -2,90 +2,76 @@
...
-    fn black() -> GdkRGBA;
-    fn copy(&self) -> GdkRGBA;
-    fn parse(&mut self, spec: &str) -> bool;
-    fn to_string(&self) -> Option<String>;
-}
+impl RGBA {
 
-impl RGBA for GdkRGBA {
-    fn white() -> GdkRGBA {
+    fn new(red: f64, green: f64, blue: f64, alpha: f64) -> RGBA {
```
EPashkin: pub fn

src/rgba.rs
```diff
@@ -2,90 +2,76 @@
...
+        ffi::GdkRGBA {
+            red: red,
+            green: green,
+            blue: blue,
+            alpha: alpha,
         }
     }
 
-    fn blue() -> GdkRGBA {
+    fn from_spec(spec: &str) -> Option<RGBA> {
```
EPashkin: pub fn

src/rgba.rs
```diff
@@ -2,90 +2,76 @@
...
         skip_assert_initialized!();
-        GdkRGBA {
-            red: 0f64,
-            green: 0f64,
-            blue: 1f64,
-            alpha: 1f64
+        unsafe {
+            let mut ffiRGBA = mem::uninitialized();
```
EPashkin: change name to `ffi_rgba` to remove warning, I copied without thought

src/rgba.rs
```diff
@@ -2,90 +2,76 @@
...
 
-    fn black() -> GdkRGBA {
-        skip_assert_initialized!();
-        GdkRGBA {
-            red: 0f64,
-            green: 0f64,
-            blue: 0f64,
-            alpha: 1f64
-        }
+    fn white() -> RGBA {
```
EPashkin: pub, others too

src/rgba.rs
Susurrus: Yeah, `new()` currently doesn't compile and I'm uncertain how to actually implement it. Should it be something like:
```
let mut rgba = RGBA::uninitialized();
rgbba.deref_mut().red = red;
...
```

That doesn't compile for me, but I'm not really sure how else to create a new `GdkRGBA` struct and embed it in the `auto::RGBA` type.

EPashkin:
```
        let ffi_rgba = ffi::GdkRGBA {
            red: red,
            green: green,
            blue: blue,
            alpha: alpha,
        };
        unsafe {
            from_glib_full(ffi::gdk_rgba_copy(&ffi_rgba))
        }
```

Susurrus:
Here's a revised version that compiles at least, though I don't know how correct it is as I don't have any code that currently uses it. I don't see much test code in the repo. Should I implement some basic tests to make sure things are in order?

EPashkin:
@Susurrus No, currently testing gtk almost impossible as test always create new thread.
Thanks.
After merge this you do for PR on gtk ? https://github.com/gtk-rs/gdk/pull/139#issuecomment-284207640
@GuillaumeGomez, It compiles with my version of `colorbutton` so 👍

src/auto/r_g_b_a.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub struct RGBA(Boxed<ffi::GdkRGBA>);
+
+    match fn {
+        copy => |ptr| ffi::gdk_rgba_copy(mut_override(ptr)),
+        free => |ptr| ffi::gdk_rgba_free(ptr),
+    }
+}
+
+impl RGBA {
+
```
GuillaumeGomez: Damn, we still generate empty line. :'(

src/rgba.rs
```diff
@@ -2,90 +2,76 @@
...
-    fn white() -> GdkRGBA;
-    fn blue() -> GdkRGBA;
-    fn green() -> GdkRGBA;
-    fn red() -> GdkRGBA;
-    fn black() -> GdkRGBA;
-    fn copy(&self) -> GdkRGBA;
-    fn parse(&mut self, spec: &str) -> bool;
-    fn to_string(&self) -> Option<String>;
-}
+impl RGBA {
```
GuillaumeGomez: Please remove this empty line.

EPashkin:
Seems gir version wrong, must be 26020e8

EPashkin:
Was wrong 26020e8 too old, please, update gir submodule to c3b402047 and regen with it

Susurrus:
@EPashkin You can actually limit the number of threads used by test according to this: https://doc.rust-lang.org/book/testing.html#testing-and-concurrency I don't know if that solves the actual problem, however.

EPashkin:
It not helps :cry:
I can't find issue, but test always create new thread even with `RUST_TEST_THREADS`, so only solution do all test in one test function, and that not good.

Susurrus:
Alright, here is hopefully a final version. Resolved the minor style changes for @GuillaumeGomez by updating gir and regenerating everything. Hopefully putting it in these 3 commits was the right move.

Susurrus:
I have been working on a GTK build, but I'm getting errors like:
```
error[E0277]: the trait bound `gdk::RGBA: glib::IsA<glib::Object>` is not satisfied
   --> src/auto/text_tag.rs:284:107
    |
284 |             gobject_ffi::g_object_set_property(self.to_glib_none().0, "foreground-rgba".to_glib_none().0, Value::from(foreground_rgba).to_glib_none().0);
    |                                                                                                           ^^^^^^^^^^^ the trait `glib::IsA<glib::Object>` is not implemented for `gdk::RGBA`
    |
    = note: required because of the requirements on the impl of `glib::value::SetValueOptional` for `gdk::RGBA`
    = note: required because of the requirements on the impl of `std::convert::From<std::option::Option<&gdk::RGBA>>` for `glib::Value`
    = note: required by `std::convert::From::from`
```

I don't think `gdk::RGBA` should be a `glib::Object`, so I'm not really certain what the correct move here is.

EPashkin:
@Susurrus, please read https://github.com/gtk-rs/gdk/pull/139#issuecomment-284207640

Susurrus:
@EPashkin Sorry I missed that. So my understanding would we'd just fix it for the current functions? My goal with this project is to get `gtk::StyleContext::get_color()` working for my project (gtk-rs/gtk#454). This function is auto-generated and just missing the RGBA type (in fact 27 functions would be supported by fixing this). Instead should I manually implement get_color instead of relying on auto-generation?

EPashkin:
If you need `get_color` now, then IMHO you need add ignoring all properties (you say 27 of it) that caused error (with comment #value gdk::RGBA).
Adding new file with manual `get_color` that hopefully will removed before next version IMHO worse but I don't against it

Susurrus:
@EPashkin Yep, I'm figuring it out now. Gonna do the 3 functions for `StyleContext` and hopefully this will eventually become automatically generated stuff.

I'm working on the GTK PR now.

Susurrus:
One problem with this implementation is that it doesn't allow easily modifying the RGBA values after creation. I'm trying to copy some existing GTK code that looks like:
```
let foreground = style_context.get_color(w.get_state_flags());
let mut background = foreground;
background.alpha *= 0.3;
```

With the current RGBA implementation you can't easily scale all values like this unless I'm missing something. Additionally is there any way to modify the individual colors after creating an `RGBA`?

Susurrus:
I just submitted by GTK PR: gtk-rs/gtk#462

EPashkin:
@Susurrus, as @GuillaumeGomez approved this PR you can merge it yourself.
Maybe github don't allow it due build failure?

EPashkin:
But as it still not merged we have time to add minor fix for better RGBA usability: `AsRef` and `AsMut` traits.

Susurrus:
Here's an updated version that adds `AsRef`, `AsMut`, and one I wanted `MulAssign`.

Susurrus:
One thing that I'm also not following, when I'm trying to use this new code, is how to use the `ContextExt::set_source_rgba` function, because if I have a widget and call `set_source_rgba()`, I get the one that takes 4 floats. Is there something I'm missing here or is that trait not implemented properly? The code I'm working with looks like:
```
let icon = gtk::DrawingArea::new();
operations_icon.connect_draw(|w, c| {
    let foreground = style_context.get_color(w.get_state_flags());
    c.set_source_rgba(&foreground);
}
```

EPashkin:
`WidgetExt::connect_draw<F: Fn(&Self, &cairo::Context)`, and cairo's has function that takes 4 floats.

EPashkin:
About this PR and GTKs.
Firstly, I made mistake when asked `AsRef`, actually I mean `Deref`.
```
impl std::ops::Deref for RGBA {
    type Target = ffi::GdkRGBA;
    
    fn deref(&self) -> &ffi::GdkRGBA {
        unsafe {
            let ptr: *const ffi::GdkRGBA = self.to_glib_none().0;
            &*ptr
        }
    }
}

impl std::ops::DerefMut for RGBA {
    fn deref_mut(&mut self) -> &mut ffi::GdkRGBA {
        unsafe {
            let ptr: *mut ffi::GdkRGBA = self.to_glib_none_mut().0;
            &mut *ptr
        }
    }
}
```
With this `MulAssign` better removed and changed to local function when needed (with comprehensible name).

EPashkin:
Second, only found this code for  `ColorChooser::add_palette`:
```
    fn add_palette(&self, orientation: Orientation, colors_per_line: i32, colors: [RGBA]) {
        unsafe {
            let mut palette = colors.iter().map(|c| local_copy(c)).collect::<Vec<_>>();
            ffi::gtk_color_chooser_add_palette(self.to_glib_none().0, orientation.to_glib(),
                colors_per_line, colors.len() as c_int, palette.as_mut_ptr() as *mut gdk_ffi::GdkRGBA)
        }
    }
} //impl<O: IsA<ColorChooser>> ColorChooserExt for O {

fn local_copy(rgba: &RGBA) -> gdk_ffi::GdkRGBA {
    let other = rgba.to_glib_none().0;
    unsafe {
        gdk_ffi::GdkRGBA {
            red: (*other).red,
            green: (*other).green,
            blue: (*other).blue,
            alpha: (*other).alpha,
        }
    }
}
```

Tested with this code in `examples\src\gtktest.rs`
```
        let color_button: ColorButton = builder.get_object("color_button").unwrap();
        let color = color_button.get_rgba();
println!("Cur: {}", color); 
        color_button.set_rgba(&color);
        let wh = RGBA::white();
        let rd = RGBA::red();
        let gr = RGBA::green();
        let bl = RGBA::blue();
println!("Green: {}", gr.green); 
        let mut gr2 = gr.clone();
        gr2.green *= 0.3;
println!("Green2: {}", gr2.green); 
        color_button.set_rgba(&wh);
        color_button.set_rgba(&gr2);
        color_button.add_palette(Orientation::Horizontal, 3, &[rd, gr, bl, wh, gr2]);
```

EPashkin:
Third, usage boxed version of RGBA looks ugly for me.
Maybe we need change approach: ignore presence `ffi::gdk_rgba_copy` and `ffi::gdk_rgba_free` and use in in `gdk::rectangle`s way:
- define `struct RGBA` with all needed `derive`
- implement `Uninitialized`, `ToGlibPtr`, `ToGlibPtrMut`, `FromGlibPtrNone`
- implement `FromGlibPtrFull` only if needed with create `RGBA` and free `ffi::GdkRGBA`

@GuillaumeGomez, @Susurrus what you think?

Susurrus:
Re your last comment @EPashkin it does seem like we're trying to cram the RGBA idea into a `Boxed` type and it's not an exact match so we're doing a lot of workarounds. I don't really understand all the reference handling and memory management that goes on behind the scenes to really have much of a comment as to the right approach. I'll defer to you guys on this.

Susurrus:
Alright, well here is a complete implementation including `Deref` and `DerefMut` and removing the `MulAssign` trait. This is a little clumsy of an interface however because if you want to modify the `RGBA` struct at all you need to `Deref` it. Here's some actual code I have:
```
    operations_icon.connect_draw(|w, c| {
                let style_context = w.get_style_context().unwrap();
                let foreground = style_context.get_color(w.get_state_flags());
                let mut background = foreground.clone();
                {
                    let background_rgba = background.deref_mut();
                    background_rgba.alpha *= 0.3;
                }
                let width = w.get_allocated_width() as f64;
                let height = w.get_allocated_height() as f64;
                let two_pi = 2.0 * std::f64::consts::PI;
                <Context as ContextExt>::set_source_rgba(&c, &background);
                c.arc(width / 2.0, height / 2.0, width.min(height) / 2.0, 0.0, two_pi);
                c.fill();
                c.move_to(width / 2.0, height / 2.0);
                <Context as ContextExt>::set_source_rgba(&c, &foreground);
                let arc_start = -std::f64::consts::FRAC_PI_2;
                let arc_end = arc_start + state.send_file_percentage as f64 / 100.0 * two_pi;
                c.arc(width / 2.0, height / 2.0, width.min(height) / 2.0, arc_start, arc_end);
                c.fill();
    });
```

That code block in the middle is so I can modify the struct and still pass it on to future functions.

I'll take a crack at implementing this instead the same way `Rectangle` does as you outlined above @EPashkin and see how that improves the ergonomics.

EPashkin:
You can write `background.alpha *= 0.3;`

EPashkin:
Thanks for trying other variant, but add new PR for it, that way we can compare

Susurrus:
@EPashkin See #140 for my alternate (and IMHO much better) implementation.

GuillaumeGomez:
I think we can close this one then?

EPashkin:
Yes, and please restart CI for https://github.com/gtk-rs/gtk/pull/462

PR #140, opened 2017-03-07
Alternate rgba implementation

Susurrus:
That takes into account all the conversation over in #139 and implements RGBA as a custom struct. The end-use usability between this and #139 is vastly improved. Basically every function supported by RGBA corresponds to an existing trait in Rust.

I've done some testing of some of the various code here as I have GTK working with these changes and my own application using both. I don't fully understand the memory management done by GTK for the `GdkRGBA` struct behind the scenes but I haven't run into crashes here.

Do note that there is one outstanding issue here and that's the error type to use for the `FromStr` trait. Since the GTK interface just uses an `Option` type but this trait uses `Result` I had to make up an error type. It looks like Glib has some error types but none of them looked appropriate. I think it's good to implement this trait but I don't think this is the best way to do the error type. Anyone have any ideas?


Susurrus:
I just realized that I didn't implement `copy` or `free` like the GTK interface has. I don't know if they're necessary, as like I said above I don't really get the memory management strategy used by GTK and how gtk-rs interfaces with it. But I'll leave it for now until I get some feedback from you guys.

src/rgba.rs
```diff
@@ -2,90 +2,155 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let string: String = unsafe {
+            from_glib_full(ffi::gdk_rgba_to_string(self.to_glib_none().0))
+        };
+        f.write_str(&string)
     }
+}
 
-    fn parse(&mut self, spec: &str) -> bool {
+impl Eq for RGBA {}
```
EPashkin: move this after PartialEq

Susurrus: I had just arranged everything alphabetically, but I'll move this if you like.

EPashkin: Actually, better move `FromStr`, `Hash`, `PartiallyEq` here. This way all xGlibPtrx will go to one place

EPashkin: At minimum better not separate Eq and Partial Eq

EPashkin:
with this Gtks `ColorChooserExt.addPalette` again looks like
```
ffi::gtk_color_chooser_add_palette(self.to_glib_none().0, orientation.to_glib(),
    colors_per_line, colors.len() as c_int, colors.as_ptr() as *mut gdk_ffi::GdkRGBA)
```
and both function of `StyleContext` need minor change to use
 `let mut rgba = gdk::RGBA::uninitialized();` that don't needed `from_glib_xx` conversion

EPashkin:
About usage `copy`, `free`: IMHO `copy` better just ignored
and maybe need add `FromGlibPtrFull` just in case.
It need call `gdk_rgba_free` after get local copy of `RGBA`

EPashkin:
also cc @GuillaumeGomez 
Seems I found way to build all gtk generated code when enabling `RGBA`.
Currently don't see way to move it to glib due conflicting implementation errors,
so gtk need additional dependence gobject-sys

```
impl glib::StaticType for RGBA {
    fn static_type() -> glib::types::Type {
        unsafe { from_glib(ffi::gdk_rgba_get_type()) }
    }
}

impl glib::value::FromValueOptional for RGBA {
    unsafe fn from_value_optional(value: &glib::Value) -> Option<Self> {
        from_glib_full(gobject_ffi::g_value_dup_boxed(value.to_glib_none().0) as *mut ffi::GdkRGBA)
    }
}

impl glib::value::SetValue for RGBA {
    unsafe fn set_value(value: &mut glib::Value, this: &Self)  {
        gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as gconstpointer)
    }
}

impl glib::value::SetValueOptional for RGBA {
    unsafe fn set_value_optional(value: &mut glib::Value, this: Option<&Self>) {
        gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as gconstpointer)
    }
}
```

GuillaumeGomez:
```rust
unsafe fn set_value(value: &mut glib::Value, this: &Self)  {
```

Should `Self` be first?

EPashkin:
This as defined in glib:
```
/// Sets a value.
pub trait SetValue: StaticType {
    unsafe fn set_value(&mut Value, &Self);
}
```
may to be same as 
```
pub trait SetValueOptional: SetValue {
    unsafe fn set_value_optional(&mut Value, Option<&Self>);
}
```
```

GuillaumeGomez:
Hum ok, too bad...

Susurrus:
So what needs to be done to wrap up this PR? From your comments it looks like:

1. Reorder trait implementations
2. Implement `FromGlibPtrFull`
3. Add glib trait impl shown in https://github.com/gtk-rs/gdk/pull/140#issuecomment-284987276

Is that it?

EPashkin:
For me it seems complete list.

Susurrus:
Alright, that should all be implemented. Local code that I have is still working correctly. I also updated gtk-rs/gtk#462 so that it's now dependent on this PR.

EPashkin:
@Susurrus, Thanks
I have minor worry about `FromStr` trait error, but prefer skip it now.

@GuillaumeGomez, what you think: this PR or #139 ?

GuillaumeGomez:
Code seems ok to me but CIs errors seem legit.

EPashkin:
Whose?

GuillaumeGomez:
check_init

Susurrus:
Whoops, forgot I left those out and meant to ask about what functions actually needed them. Good catch!

So I've pushed a new version with fixes.

Susurrus:
I pushed another version that fixes a minor style issue and also exports the `RgbaParseError` type.

EPashkin:
Ping

GuillaumeGomez:
Oups, forgot about it... Thanks!

PR #145, opened 2017-03-17
Finish implementing gdk::Atom

Susurrus:
I think this is all that's necessary?


EPashkin:
Seems you need define `GlibPtrDefault` to fulfill need of https://github.com/gtk-rs/glib/blob/master/src/translate.rs#L967
```
impl GlibPtrDefault for Atom {
    type GlibType = *mut ffi::GdkAtom;
}
```

Susurrus:
The trait bounds aren't satisfied even with the addition of a `GlibPtrDefault` impl for `Atom`. The compiler errors are complaining about a lack of this trait for `Vec<Atom>`, but I don't think that should actually be implemented. Is there something else missing here?

EPashkin:
1. make `FromGlibPtrNone` and `FromGlibPtrFull` works from pointers not from value
```rust
//src/atom.rs
impl FromGlibPtrNone<*mut ffi::GdkAtom> for Atom {
    #[inline]
    unsafe fn from_glib_none(ptr: *mut ffi::GdkAtom) -> Atom { Atom(*ptr) }
}
```
2. Fix value atom errors
```rust
//src/event_owner_change.rs
     pub fn get_selection(&self) -> ::Atom {
        unsafe { from_glib_none(self.as_ref().selection as *mut _) }
     }
```
3. remove manual display.rs

Susurrus:
I think this corrects the remaining issues.

src/auto/display.rs
```diff
@@ -257,9 +260,11 @@ impl Display {
         }
     }
 
-    //pub fn store_clipboard(&self, clipboard_window: &Window, time_: u32, targets: /*Ignored*/&[&Atom], n_targets: i32) {
-    //    unsafe { TODO: call ffi::gdk_display_store_clipboard() }
-    //}
+    pub fn store_clipboard(&self, clipboard_window: &Window, time_: u32, targets: &[&Atom], n_targets: i32) {
+        unsafe {
+            ffi::gdk_display_store_clipboard(self.to_glib_none().0, clipboard_window.to_glib_none().0, time_, targets.to_glib_none().0, n_targets);
```
EPashkin: This function need tested with at minimum 2 element array.
Also about usage: not sure this or `targets: &[Atom]` better.

src/auto/device.rs
```diff
@@ -206,9 +207,11 @@ impl Device {
         }
     }
 
-    //pub fn list_axes(&self) -> /*Ignored*/Vec<Atom> {
-    //    unsafe { TODO: call ffi::gdk_device_list_axes() }
-    //}
+    pub fn list_axes(&self) -> Vec<Atom> {
+        unsafe {
+            FromGlibPtrContainer::from_glib_container(ffi::gdk_device_list_axes(self.to_glib_none().0))
```
EPashkin: This function need be tested too.
For my default display it returns 2 devices and this function return empty vector, so I can't test it
```rust
let display = gdk::Display::get_default().unwrap();
let devices = display.list_devices();
println!("{}", devices.len());

for device in devices {
    println!("name: {:?}", device.get_name());
    let axes = device.list_axes();
    println!("{}", axes.len());
}
```

Susurrus: Indeed, I'm also getting 0's for `list_axes` for my two devices. Do you know what a `GdkDevice` is in practice, and output device or an input device? I can't glean much information from various internet searches.

src/auto/display.rs
Susurrus: I think we want this to be `&[Atom]` since it's an input. The natural way to generate this will be to create a `Vec<Atom>` and pass a slice to the function. If the type is a slice of references, it's going to take an extra conversion to use this in the common case I think. It also shouldn't have the `target_length` argument, so I'll need to manually implement this.

Susurrus:
Pushed a version that ignores the `store_clipboard` function for now and squash into 2 commits.

src/auto/display.rs
EPashkin: Then `src/display.rs` need be restored with it

Susurrus:
@EPashkin Is there anything else left to fix with this or are you waiting for gtk-rs/gtk#472 is in solid shape before you want to merge this?

EPashkin:
As I don't understand what `gdk_display_store_clipboard` must do current version good for me and can be merged.
@GuillaumeGomez What you think about merging this now?

GuillaumeGomez:
I didn't see anything blocking so if @EPashkin says ok, it's good to get merged. Thanks!

PR #147, opened 2017-03-27
Removed missed manual duplicate modifier_type

EPashkin:
Duplicate of generated https://github.com/gtk-rs/gdk/blob/master/src/auto/flags.rs#L184-L219

### DIFF
--- a/src/enums.rs
+++ b/src/enums.rs
@@ -3,30 +3,6 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 #[allow(non_upper_case_globals)]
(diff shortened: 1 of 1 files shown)


EPashkin:
@GuillaumeGomez Maybe also better move keys to it module and remove `src/enums.rs`?
Also it need fix in examples first

GuillaumeGomez:
I don't have strong opinion on this. If what you propose is better, then no problem. ;)

EPashkin:
@GuillaumeGomez I also not sure about need of move keys, so currently leave it in enums.

GuillaumeGomez:
Fine by me.

PR #148, opened 2017-03-27
Add missing impls for Atom

Susurrus:
Fixes necessary for gtk-rs/gtk#478

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -44,6 +44,16 @@ impl FromGlibPtrFull<*mut ffi::GdkAtom> for Atom {
     unsafe fn from_glib_full(_: *mut ffi::GdkAtom) -> Atom { unimplemented!() }
 }
 
+impl FromGlibPtrNone<ffi::GdkAtom> for Atom {
+    #[inline]
+    unsafe fn from_glib_none(ptr: ffi::GdkAtom) -> Atom { Atom(ptr) }
+}
+
+impl FromGlibPtrFull<ffi::GdkAtom> for Atom {
+    #[inline]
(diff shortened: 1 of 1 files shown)


EPashkin:
👍 without it function returning Atom not work

Susurrus:
As part of this Atom work we're also missing all of the `Atom`s specified on the [Selections page](https://developer.gnome.org/gdk3/stable/gdk3-Selections.html). I don't even see them in gdk-sys though, so should I file a separate issue on the `sys` repo or how should I go about adding these?

EPashkin:
May be it can be done as https://github.com/gtk-rs/gdk/blob/master/src/atom.rs#L11 but it need checked.
Also we don't have any of selection functions, so maybe these const need go to separate `manual/selection.rs` with adding some functions.
Also seems these selection functions X11 related and need be hidden under `#[cfg(target_os = "linux")]` but may be not, currently I don't ready fo check it.
Maybe @GuillaumeGomez know more about it?
I not against add constants somewhere if its really needed, but not sure if it need be added in this PR as it stall other PRs.

Susurrus:
Right now the only thing that would be nice would be the constants. I'll add those as a separate issue and we can keep this PR focused on just this small part!

GuillaumeGomez:
@EPashkin: If it's X11 specific, then we need to make conditional compilation for it like we did in some other parts as well.

EPashkin:
@GuillaumeGomez X11 specific part moved to separate issue #149 (if it really specific)

GuillaumeGomez:
Ah my bad. Then it's all good for me as well. Thanks @Susurrus!

PR #150, opened 2017-03-30
Fixed Atom container conversion

EPashkin:
Fixed errors found in https://github.com/gtk-rs/examples/pull/119

Removed unneeded FromGlibPtrXXX
Added Atom to usize conversion

@GuillaumeGomez, @Susurrus can you test `Atom::value`?
For windows it return small numbers like 31 not sure about linux and mac
```rust
let atom = gdk::Atom::intern("STRING");
println!("atom:{}", atom.name());
println!("{}", unsafe{atom.value()});
```

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -19,10 +19,14 @@ impl Atom {
     pub fn name(&self) -> String {
         unsafe { from_glib_full(ffi::gdk_atom_name(self.0)) }
     }
+
+    pub unsafe fn value(&self) -> usize {
+        self.0 as usize
+    }
 }
 
 impl GlibPtrDefault for Atom {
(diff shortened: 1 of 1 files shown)


Susurrus:
Prints
```
atom:STRING
31
```
for me on Linux x64.

Thanks @EPashkin for continuing to dig into this, it looks like this fixes the `target_list` problem we had too. All that's left for gtk-rs/examples#119 is figuring out how to get the `drag-data-received` signal to actually be triggered.

EPashkin:
cc @GuillaumeGomez, what you think about this fix?

GuillaumeGomez:
Less indirection so I approve!

PR #162, opened 2017-05-15
Regenerate autogenerated bindings with latest GIR

sdroege:


sdroege:
Sorry, push-forced pango in here. It's all confused now so I created a new pull request

PR #170, opened 2017-06-12
Add remaining GdkAtom constants

### LINKED ISSUE #120: Missing clipboard constants
As far as I can tell, this crate is missing bindings to the `GDK_SELECTION_CLIPBOARD`, `GDK_SELECTION_PRIMARY`, and `GDK_SELECTION_SECONDARY` constants. These constants are necessary to implement cut/copy/paste with GTK.

Susurrus:
Closes #120.


GuillaumeGomez:
:+1:

EPashkin:
Atom::None was reexported https://github.com/gtk-rs/gdk/blob/master/src/lib.rs#L74
and `mod atom` currently not public

Susurrus:
@EPashkin How would you like these exported? Everything as `Atom::*` or should I re-export all these constants in the root?

EPashkin:
As atom::None exported by different name I not sure :(

EPashkin:
IMHO better just move all consts from atom.rs to lib.rs

Susurrus:
What should be done about `Atom::None` then? And what should the final public exported names be?

EPashkin:
IMHO ATOM_NONE better stay as is, and for other current names is good

Susurrus:
If I move those constants out of `atom.rs` then I can't use the constructor. Should I just re-export them then or how should I organize this?

EPashkin:
Your right. Then we have only one option: direct reexport under same name.
I don't see how it can be organized differently.

Susurrus:
How about this?

EPashkin:
👍 
Thanks, @Susurrus

Susurrus:
This need anything else to be merged? Tests pass.

GuillaumeGomez:
Just to ping me. ;)

Thanks!

PR #180, opened 2017-07-19
Fix transfer full conversion of Atom arrays

sdroege:
Instead of the original pointer, the pointer at the end of the array was
being freed.

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -100,15 +100,14 @@ impl FromGlibPtrFull<ffi::GdkAtom> for Atom {
 }
 
 impl FromGlibContainerAsVec<ffi::GdkAtom, *mut ffi::GdkAtom> for Atom {
-    unsafe fn from_glib_none_num_as_vec(mut ptr: *mut ffi::GdkAtom, num: usize) -> Vec<Self> {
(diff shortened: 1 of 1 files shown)


EPashkin:
As no g_free called, this PR IMHO unneeded

GuillaumeGomez:
@EPashkin: The point is to remove the unnecessary mutability. :)

EPashkin:
As I know `mut ptr: *mut ffi::GdkAtom` is syntax sugar for internal mutability and converted by compiler to 
```rust
unsafe fn from_glib_none_num_as_vec(ptr: *mut ffi::GdkAtom, num: usize) -> Vec<Self> {
let mut ptr = ptr;
...
}
```
and `mut ptr` even don't appeared in docs.
So I still think that this already was good enough.

sdroege:
Sorry, forgot "git commit --amend". There actually is a g_free() here.

sdroege:
For the other function change, see https://github.com/gtk-rs/glib/pull/200#discussion_r128202171

Let's use the compiler for preventing mistakes instead of trying to be too clever

GuillaumeGomez:
Absolutely. Thanks!

PR #204, opened 2018-02-03
Regen to lattest version

EPashkin:
Apply [last gir-files update](https://github.com/gtk-rs/gir-files/pull/14)

cc @GuillaumeGomez , @sdroege 

Only addition in auto/flags.rs


GuillaumeGomez:
Thanks!

PR #221, opened 2018-06-24
Fix clippy warnings

EPashkin:
cc @GuillaumeGomez

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,6 +3,10 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 #![allow(deprecated)]
+#![cfg_attr(feature = "cargo-clippy", allow(cast_ptr_alignment))]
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -3,6 +3,10 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 #![allow(deprecated)]
+#![cfg_attr(feature = "cargo-clippy", allow(cast_ptr_alignment))]
+#![cfg_attr(feature = "cargo-clippy", allow(transmute_int_to_char))]
+#![cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ptr))]
+#![cfg_attr(feature = "cargo-clippy", allow(trivially_copy_pass_by_ref))]
```
EPashkin: Many warnings about this and &Atom in generated code, so I fixed only atom.rs and ignored others

GuillaumeGomez: Wow, that's a lot indeed!

GuillaumeGomez:
Thanks!

PR #223, opened 2018-06-30
Add missing EventMotion function

### LINKED ISSUE #222: Bindings for gdk_event_request_motions() missing
GDK documentation: https://developer.gnome.org/gdk3/stable/gdk3-Events.html#gdk-event-request-motions

GTK 3.8 and onwards will always compress motion events, unless gdk_window_set_event_compression() is used to disable compression. gdk_event_request_motions() is the recommended way of decompressing the events in a handler, but it seems that the binding for it is missing. An alternative is the gdk_window_get_device_position() function, which does have a binding, but EventMotion doesn't seem to provide a .get_device() method on the Rust side.

GuillaumeGomez:
Fixes #222.

cc @sdroege @EPashkin

### DIFF
--- a/src/event_motion.rs
+++ b/src/event_motion.rs
@@ -25,4 +25,8 @@ impl EventMotion {
     pub fn get_time(&self) -> u32 {
         self.as_ref().time
     }
+
+    pub fn request_motions(&self) {
+        unsafe { ffi::gdk_event_request_motions(self.as_ref()) }
+    }
 }


EPashkin:
PR looks good itself, but strange, appveyor failing on gio.

GuillaumeGomez:
That's super weird. O.o

sdroege:
Unrelated to this PR, it's from the subprocess PR. For some reason it goes into the Unix code paths on Windows, so something needs to be changed there

This PR is fine and can be merged

GuillaumeGomez:
Yes, I'm looking at the gio code right now trying to understand what's going on.

Thanks for reviews!

PR #224, opened 2018-07-01
New types

GuillaumeGomez:
cc @sdroege @EPashkin


src/lib.rs
```diff
@@ -130,3 +130,13 @@ pub use self::keys::{
     keyval_name,
     keyval_to_unicode
 };
+
+pub fn pango_layout_get_clip_region(layout: &pango::Layout, x_origin: i32, y_origin: i32, index_ranges: &[i32]) -> Option<cairo::Region> {
+    use glib::translate::from_glib_full;
+    use glib::translate::ToGlibPtr;
+
+    assert_initialized_main_thread!();
```
EPashkin: Assert need be on first line.
Also IMHO better add separate "functions.rs" now.

Gir.toml
```diff
@@ -318,6 +325,9 @@ status = "generate"
     name = "new"
     ignore = true
     [[object.function]]
+    pattern = "[gs]et_background_pattern"
+    ignore = true
```
EPashkin: Please add comment, why it ignored

src/auto/functions.rs
```diff
@@ -237,11 +238,7 @@ pub fn pango_context_get_for_screen(screen: &Screen) -> Option<pango::Context> {
     }
 }
 
-//pub fn pango_layout_get_clip_region(layout: /*Ignored*/&pango::Layout, x_origin: i32, y_origin: i32, index_ranges: i32, n_ranges: i32) -> Option<cairo::Region> {
-//    unsafe { TODO: call ffi::gdk_pango_layout_get_clip_region() }
-//}
-
-//pub fn pango_layout_line_get_clip_region(line: /*Ignored*/&pango::LayoutLine, x_origin: i32, y_origin: i32, index_ranges: &[i32], n_ranges: i32) -> Option<cairo::Region> {
+//pub fn pango_layout_line_get_clip_region(line: &pango::LayoutLine, x_origin: i32, y_origin: i32, index_ranges: &[i32], n_ranges: i32) -> Option<cairo::Region> {
```
EPashkin: This function seems also need manual implementation

sdroege: Why is it not in the commented output why a manual impl is needed?

But the `index_ranges` and `n_ranges` need some manual work here. The `n_ranges` is half of the elements in `index_ranges`, and `index_ranges` is `array of byte indexes into the layout, where even members of array are start indexes and odd elements are end indexes.`

src/auto/functions.rs
```diff
@@ -413,9 +410,12 @@ pub fn set_show_events(show_events: bool) {
     }
 }
 
-//pub fn setting_get(name: &str, value: /*Ignored*/&mut glib::Value) -> bool {
-//    unsafe { TODO: call ffi::gdk_setting_get() }
-//}
+pub fn setting_get(name: &str, value: &mut glib::Value) -> bool {
```
sdroege: Does the `GValue` have to be initialized with the correct type before calling? Or all zeroes?

You probably want to ensure either inside the function and return a `Option<Value>`

src/auto/screen.rs
```diff
@@ -274,9 +274,11 @@ impl<O: IsA<Screen> + IsA<glib::object::Object>> ScreenExt for O {
         }
     }
 
-    //fn get_setting(&self, name: &str, value: /*Ignored*/&mut glib::Value) -> bool {
-    //    unsafe { TODO: call ffi::gdk_screen_get_setting() }
-    //}
+    fn get_setting(&self, name: &str, value: &mut glib::Value) -> bool {
```
sdroege: Same as above

src/lib.rs
sdroege: Ah you manually implemented it. See my comments here https://github.com/gtk-rs/gdk/pull/224/files#r199389013

sdroege: You probably want the array to be an array of tuples

Gir.toml
sdroege: Especially if it is not manually implemented either

src/lib.rs
GuillaumeGomez: I don't know how tuples are represented in memory... If it's contiguous values, then it'll easy but otherwise it'll require a copy, not sure we want that.

Gir.toml
GuillaumeGomez: Deprecated. I'll add the comment.

src/auto/functions.rs
GuillaumeGomez: Oh true! I went too fast with this one...

Gir.toml
EPashkin: It deprecated in 3.22 https://developer.gnome.org/gdk3/stable/gdk3-Windows.html#gdk-window-get-background-pattern, so it can be used even in max available on CI 3.18.
So this not good reason to ignore.

GuillaumeGomez: Damn. You're hard in business. :stuck_out_tongue:

src/lib.rs
sdroege: Currently it would work but there is no guarantee I believe. The problem otherwise is that you have an easy to misuse API

sdroege: Maybe a tuple struct with packed representation would be best here. Ugly ugly C API

GuillaumeGomez: Talk about dirty hacks...

GuillaumeGomez:
Isn't it supposed to be an enum? I must have missed something... But anyway, I already tried to box it without succeeding. I'll try again tomorrow.

EPashkin:
Pattern looks like enum for me https://github.com/gtk-rs/cairo/blob/master/src/patterns.rs#L48-L55 :wink:
But error sure from trait.

sdroege:
Indeed, I misread. That looks all as if it has a static size. I wonder also where the `'static` in the compiler error comes from, that might be a hint

GuillaumeGomez:
*__FINALLY__* fixed the super ultra weird bug. Talk about weird stuff...

sdroege:
> FINALLY fixed the super ultra weird bug. Talk about weird stuff...

What was it exactly? :)

GuillaumeGomez:
rustc bug apparently.

src/window.rs
```diff
@@ -139,6 +139,10 @@ pub trait WindowExtManual {
     fn offscreen_window_get_surface(&self) -> Option<Surface>;
 
     fn get_pixbuf(&self, src_x: i32, src_y: i32, width: i32, height: i32) -> Option<gdk_pixbuf::Pixbuf>;
+
+    fn get_background_pattern(&self) -> Option<cairo::Pattern>;
```
EPashkin: Information about deprecation is lost

sdroege:
> rustc bug apparently

Do you have some more details? :)

GuillaumeGomez:
> Do you have some more details? :)

Not at all. I updated the compiler and it worked.

sdroege:
> Not at all. I updated the compiler and it worked.

From which to which version?

GuillaumeGomez:
The 7th July nightly from the one we had on CI I assume.

sdroege:
> The 7th July nightly from the one we had on CI I assume.

So it was only a transient problem with nightly, stable/beta worked all the time? Good

GuillaumeGomez:
Apparently? I'm very surprised such a bug existed in the first place...

EPashkin:
Then manual `get_background_pattern` non needed?

EPashkin:
It still needed as `cairo::Pattern` don't have "from_glib_XX" :(

GuillaumeGomez:
Yep. I merge then!

PR #227, opened 2018-07-09
Add tests

GuillaumeGomez:
Work in progress but at least you can have a look first. I'll improve the imports check a bit. Remains to do:

* [x] crates import check
* [ ] Gir.toml check
* [ ] Re-exports check
* [ ] Imports check
  * [x] normal cases (`use` without `{}`)
  * [ ] not so normal cases (`use` with `{}`)

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -3,10 +3,10 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 use ffi;
-use glib_ffi;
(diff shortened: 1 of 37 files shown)


EPashkin:
@GuillaumeGomez Sorry, but IMHO all of this is better leave to rustfmt.

GuillaumeGomez:
`rustfmt` can do it? If so, I'll just make the test for `Gir.toml`.

PR #231, opened 2018-07-17
Add missing get_device methods

### LINKED ISSUE #230: Events are missing properties
Many event objects are missing properties in rust compared to the gdk implementation.
Example: [gdk motion event](https://developer.gnome.org/gdk3/stable/gdk3-Event-Structures.html#GdkEventMotion) has a device property and [gtk-rs motion event](http://gtk-rs.org/docs/gdk/struct.EventMotion.html) does not.
Were these properties left out for a specific reason?

GuillaumeGomez:
Fixes #230.

cc @sdroege @EPashkin

### DIFF
--- a/src/event_button.rs
+++ b/src/event_button.rs
@@ -29,4 +29,8 @@ impl EventButton {
     pub fn get_button(&self) -> u32 {
         self.as_ref().button
     }
+
+    pub fn get_device(&self) -> Option<::Device> {
+        unsafe { from_glib_none(self.as_ref().device) }
+    }
 }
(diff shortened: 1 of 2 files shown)


EPashkin:
@GuillaumeGomez  Code look good, Thanks. 👍 

@sdroege There again problem with futures on nightly, seems its updated.

sdroege:
Looks good

GuillaumeGomez:
Then I merge but weird failure...

sdroege:
> Then I merge but weird failure...

I can reproduce it here but it seems to be first of all a bug in nightly. It complains to add `#![feature(proc_macro)]` but that already is there.

sdroege:
See https://github.com/rust-lang-nursery/futures-rs/issues/1096

PR #232, opened 2018-07-18
Add missing getters

GuillaumeGomez:
cc @EPashkin @sdroege

### DIFF
--- a/src/event_button.rs
+++ b/src/event_button.rs
@@ -33,4 +33,20 @@ impl EventButton {
     pub fn get_device(&self) -> Option<::Device> {
         unsafe { from_glib_none(self.as_ref().device) }
     }
+
+    pub fn get_axes(&self) -> Option<(f64, f64)> {
+        let axes = self.as_ref().axes;
+
+        if axes.is_null() {
+            None
+        } else {
+            unsafe { Some((*axes, *axes.offset(1))) }
+        }
+    }
+
+    pub fn get_root(&self) -> (f64, f64) {
+        let x_root = self.as_ref().x_root;
+        let y_root = self.as_ref().y_root;
+        (x_root, y_root)
+    }
 }
--- a/src/event_key.rs
+++ b/src/event_key.rs
@@ -38,5 +38,8 @@ impl EventKey {
         self.as_ref().group
     }
 
-    // TODO: add getter for is_modifier
+    // TODO: to be uncommented when the field will be generated
+    // pub fn get_is_modifier(&self) -> bool {
+    //     from_glib(self.as_ref().is_modifier)
+    // }
 }
--- a/src/event_scroll.rs
+++ b/src/event_scroll.rs
@@ -45,4 +45,9 @@ impl EventScroll {
         let dy = self.as_ref().delta_y;
         (dx, dy)
     }
+
+    // TODO: to be uncommented when the field will be generated
+    // pub fn get_is_stop(&self) -> bool {
+    //     from_glib(self.as_ref().is_stop)
+    // }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,8 +66,11 @@ mod event_proximity;
 mod event_scroll;
 mod event_selection;
 mod event_setting;
+#[cfg(any(feature = "v3_4", feature = "dox"))]
 mod event_touch;
+#[cfg(any(feature = "v3_18", feature = "dox"))]
 mod event_touchpad_pinch;
+#[cfg(any(feature = "v3_18", feature = "dox"))]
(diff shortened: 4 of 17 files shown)


src/event_button.rs
```diff
@@ -33,4 +33,14 @@ impl EventButton {
...
+
+    pub fn get_axes(&self) -> Option<(f64, f64)> {
+        let axes = self.as_ref().axes;
+
+        if axes.is_null() {
+            None
+        } else {
+            unsafe { Some((*axes, *axes.offset(1))) }
+        }
+    }
```
sdroege: Please check the docs and add *all* the missing getters while we're at it :)

There's at least `x_root` and `y_root` still missing here and maybe other things for the other events

GuillaumeGomez: Damn, I thought it was implemented in `Event`... I'll need to make a bigger check.

GuillaumeGomez:
Added a *lot* of missing stuff.

src/lib.rs
```diff
@@ -55,12 +55,20 @@ mod event_grab_broken;
...
+#[cfg(any(feature = "v3_22", feature = "dox"))]
+mod event_pad_group_mode;
 mod event_property;
 mod event_proximity;
 mod event_scroll;
 mod event_selection;
 mod event_setting;
 mod event_touch;
+mod event_touchpad_pinch;
+mod event_touchpad_swipe;
```
sdroege: These two exist since 3.18

src/lib.rs
```diff
@@ -55,12 +55,20 @@ mod event_grab_broken;
...
+#[cfg(any(feature = "v3_22", feature = "dox"))]
+mod event_pad_button;
+#[cfg(any(feature = "v3_22", feature = "dox"))]
+mod event_pad_group_mode;
 mod event_property;
 mod event_proximity;
 mod event_scroll;
 mod event_selection;
 mod event_setting;
 mod event_touch;
```
sdroege: `GDK_TOUCH_BEGIN/END/CANCEL/UPDATE` exist since 3.4, not sure how this worked without feature flag before?

src/lib.rs
```diff
@@ -111,12 +119,20 @@ pub use event_grab_broken::EventGrabBroken;
...
+#[cfg(any(feature = "v3_22", feature = "dox"))]
+pub use event_pad_group_mode::EventPadGroupMode;
 pub use event_property::EventProperty;
 pub use event_proximity::EventProximity;
 pub use event_scroll::EventScroll;
 pub use event_selection::EventSelection;
 pub use event_setting::EventSetting;
 pub use event_touch::EventTouch;
+pub use event_touchpad_pinch::EventTouchpadPinch;
+pub use event_touchpad_swipe::EventTouchpadSwipe;
```
sdroege: 3.18

src/event_key.rs
```diff
@@ -38,5 +38,8 @@ impl EventKey {
         self.as_ref().group
     }
 
-    // TODO: add getter for is_modifier
+    // TODO: to be uncommented when the field will be generated
+    // pub fn get_is_modifier(&self) -> bool {
+    //     from_glib(self.as_ref().is_modifier)
+    // }
```
sdroege: Bitfields /o\

As the whole `GdkEvent` union is zero-allocated, you can
```
    let ptr = (&self.as_ref().group as *const u8).offset(1);
    from_glib(*ptr)
```

src/event_scroll.rs
```diff
@@ -45,4 +45,9 @@ impl EventScroll {
         let dy = self.as_ref().delta_y;
         (dx, dy)
     }
+
+    // TODO: to be uncommented when the field will be generated
+    // pub fn get_is_stop(&self) -> bool {
+    //     from_glib(self.as_ref().is_stop)
```
sdroege: See https://github.com/gtk-rs/gdk/pull/232/files#r203947133, same story

sdroege:
Looks good

src/event_key.rs
GuillaumeGomez: I'd rather not. I prefer to just wait for the generation instead.

GuillaumeGomez:
Then I merge!

src/event_key.rs
sdroege: We can probably wait years for that, it's far from trivial to get right for the general case

GuillaumeGomez: I'm a patient man. :)

sdroege: The users might not be :P

GuillaumeGomez: I'll think about it when I'll hear their complains then. :p

PR #233, opened 2018-07-20
Add not_bound rule

GuillaumeGomez:


GuillaumeGomez:
Since it's just a Makefile rule, no need to wait for tests.

Issue #129, opened 2016-09-29
add event_handler_set wrapper


aeosynth:
it's not obvious to me how to use the `gdk_sys` version, so i'd like a higher level wrapper.

PR #238, opened 2018-09-04
Implement #129: use wrapper callbacks to convert arguments

dhardy:
Since there is no existing use of callbacks in GDK I had to improvise. I'm not certain if this is the ideal approach (there's the overhead of another function call which *probably* won't be optimised away, at least without full LTO), and it's not as simple as I'd like, but it may still be the best we can do.

Alternatively we could simplify by dropping the `data` and possibly `notify` arguments; I'm not sure how useful they are, though presumably a binding lib should just follow the underlying API as closely as possible.

I haven't tested it.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -56,32 +56,24 @@ impl Event {
 
     /// Set the event handler.
     /// 
-    /// The callback `func` is called for each event. If `None`, event handling
-    /// is disabled.
-    /// 
-    /// The `data` pointer is passed to `func` and `notify` when called.
-    /// 
-    /// Should another event handler be set in the future, `notify` will be
-    /// called before doing so, if not `None`.
-    pub fn set_handler<T>(
-        func: Option<&'static Fn(Event, Option<*mut T>)>,
-        data: Option<*mut T>,
-        notify: Option<&'static Fn(Option<*mut T>)>)
-    {
-        let data = data.unwrap_or(ptr::null_mut()) as *mut c_void;
-        unsafe{
-            CB_EVENT_HANDLER = mem::transmute::<
-                Option<&'static Fn(Event, Option<*mut T>)>,
-                Option<&'static Fn(Event, Option<*mut ()>)>
-                >(func);
-            CB_EVENT_HANDLER_NOTIFY = mem::transmute::<
-                Option<&'static Fn(Option<*mut T>)>,
-                Option<&'static Fn(Option<*mut ()>)>
-                >(notify);
-            ffi::gdk_event_handler_set(
-                Some(event_handler),
-                data,
-                Some(event_handler_notify))
+    /// The callback `handler` is called for each event. If `None`, event
+    /// handling is disabled.
+    pub fn set_handler<F: Fn(&mut Event) + Send +'static>(handler: Option<F>) {
+        if let Some(handler) = handler {
+            // allocate and convert to target type
+            // double box to reduce a fat pointer to a simple pointer
+            let boxed: Box<Box<Fn(&mut Event) + Send + 'static>> = Box::new(Box::new(handler));
+            let ptr: *mut c_void = Box::into_raw(boxed) as *mut _;
+            unsafe{
+                ffi::gdk_event_handler_set(
+                    Some(event_handler_trampoline),
+                    ptr,
+                    Some(event_handler_destroy))
+            }
+        } else {
+            unsafe{
+                ffi::gdk_event_handler_set(None, ptr::null_mut(), None)
+            }
         }
     }
 
@@ -454,22 +446,19 @@ macro_rules! event_subtype {
     }
 }
 
-static mut CB_EVENT_HANDLER: Option<&Fn(Event, Option<*mut ()>)> = None;
-static mut CB_EVENT_HANDLER_NOTIFY: Option<&Fn(Option<*mut ()>)> = None;
-
-extern "C" fn event_handler(event: *mut ffi::GdkEvent, p: glib_ffi::gpointer) {
-    let cb = unsafe {
-        if let Some(cb) = CB_EVENT_HANDLER { cb } else { return; }
-    };
-    let event = unsafe { from_glib_none(event) };
-    let p = if p == ptr::null_mut() { None } else { Some(p as *mut ()) };
-    (cb)(event, p);
+extern "C" fn event_handler_trampoline(event: *mut ffi::GdkEvent, ptr: glib_ffi::gpointer) {
+    let mut event = unsafe { from_glib_none(event) };
+    if ptr != ptr::null_mut() {
+        let raw: *mut Box<dyn Fn(&mut Event) + Send +'static> = ptr as _;
+        let f: &(dyn Fn(&mut Event) + Send +'static) = unsafe { &**raw };
+        f(&mut event)
+    }
 }
 
-extern "C" fn event_handler_notify(p: glib_ffi::gpointer) {
-    let cb = unsafe {
-        if let Some(cb) = CB_EVENT_HANDLER_NOTIFY { cb } else { return; }
-    };
-    let p = if p == ptr::null_mut() { None } else { Some(p as *mut ()) };
-    (cb)(p);
+extern "C" fn event_handler_destroy(ptr: glib_ffi::gpointer) {
+    if ptr != ptr::null_mut() {
+        // convert back to Box and free
+        let raw: *mut Box<dyn Fn(&mut Event) + Send +'static> = ptr as _;
+        let _boxed: Box<Box<dyn Fn(&mut Event) + Send +'static>> = unsafe { Box::from_raw(raw) };
+    }
 }
--- a/src/event.rs
+++ b/src/event.rs
@@ -2,6 +2,7 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
+use libc::c_void;
 use glib::translate::*;
 use ffi;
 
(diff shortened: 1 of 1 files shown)


src/event.rs
```diff
@@ -53,6 +54,37 @@ impl Event {
...
+    /// The `data` pointer is passed to `func` and `notify` when called.
+    /// 
+    /// Should another event handler be set in the future, `notify` will be
+    /// called before doing so, if not `None`.
+    pub fn set_handler<T>(
+        func: Option<&'static Fn(Event, Option<*mut T>)>,
+        data: Option<*mut T>,
+        notify: Option<&'static Fn(Option<*mut T>)>)
+    {
+        let data = data.unwrap_or(ptr::null_mut()) as *mut c_void;
```
EPashkin: This "static" function need macro like `assert_initialized_main_thread!();` in first line

src/event.rs
```diff
@@ -53,6 +54,37 @@ impl Event {
...
+    /// 
+    /// The callback `func` is called for each event. If `None`, event handling
+    /// is disabled.
+    /// 
+    /// The `data` pointer is passed to `func` and `notify` when called.
+    /// 
+    /// Should another event handler be set in the future, `notify` will be
+    /// called before doing so, if not `None`.
+    pub fn set_handler<T>(
+        func: Option<&'static Fn(Event, Option<*mut T>)>,
```
EPashkin: Second parameter in closure IMHO wrong, and don't need in rust closures at all, better always pass `data=null_ptr`

EPashkin:
Also usually we use `destroy` to "safe" free closure memory,
see https://github.com/gtk-rs/glib/blob/master/src/signal.rs#L51 as example.

GuillaumeGomez:
I recommend you to see how we handle callbacks with signals. It should help you to fix the last parts.

Also, I agree with @EPashkin on the `unsafe` part. This seems like not the good way to do this. Having it is nice but I'd prefer to let people know that they're using it at they own risk. (if you have an opinion on this @sdroege?)

sdroege:
Just some comments here

- First of all, how would you use this and why? Can you give a use-case and example? :) This only seems useful if implementing a new toolkit on top of GDK that is not GTK?

- What others said already: The closure should be taken be value and not reference, it should take only the event as argument and the lifetime of the closure should be managed via the destroy notify. Look at how we do callbacks (e.g. for signals or other things) elsewhere. I don't see why this is not possible here. Everything else is simply unsafe and you can as well use the bindings from gdk-sys then, this would not add any advantage over that here. In short: you pass the closure via the data pointer, you have a trampoline for the actual function pointer that extracts the closure from the data pointer given to the trampoline function via argument, you destroy the closure via the destroy notify. The function signature should be `set_handler<F: Fn(&gdk::Event)>(handler: Option<F>)`

dhardy:
There's a problem:
```rust
extern "C" fn event_handler_destroy(ptr: glib_ffi::gpointer) {
    if ptr != ptr::null_mut() {
        // convert back to Box and free
        let raw: *mut (dyn Fn(&Event) + Send + Sync + 'static) = ptr as _;
        unsafe{
            let boxed: Box<dyn Fn(&Event) + Send + Sync + 'static> = Box::from_raw(raw);
        }
    }
}
```
Error:
```
error[E0277]: expected a `std::ops::Fn<(&event::Event,)>` closure, found `libc::c_void`
   --> src/event.rs:460:66
    |
460 |         let raw: *mut (dyn Fn(&Event) + Send + Sync + 'static) = ptr as _;
    |                                                                  ^^^ expected an `Fn<(&event::Event,)>` closure, found `libc::c_void`
    |
    = help: the trait `for<'r> std::ops::Fn<(&'r event::Event,)>` is not implemented for `libc::c_void`
    = note: required for the cast to the object type `(dyn for<'r> std::ops::Fn(&'r event::Event) + std::marker::Send + std::marker::Sync + 'static)`
```
The error message doesn't make much sense, but I suspect the error is because a `*mut dyn Fn(..)` is a fat pointer while `*mut c_void` is not.

Potential solutions:
1. accept a function pointer instead (probably not a fat pointer; shouldn't need allocation): `fn(&Event)`
2. double-allocate (sounds stupid but should be viable and I see this used elsewhere)
3. use static memory to store the closure (which isn't an issue since GDK stores the pointer it receives in static memory anyway)

dhardy:
Which is one of the solutions I listed, but decidedly indirect. Very well.

sdroege:
Sorry, I forgot to write something about your 3 options above.

> 1. accept a function pointer instead (probably not a fat pointer; shouldn't need allocation): fn(&Event)

This would work but would limit functionality quite a bit as you can't have any state in your function (other than using global state)

> 2. double-allocate (sounds stupid but should be viable and I see this used elsewhere)

As this not something you're setting very often, the allocation doesn't really matter and in the greater scheme of things it's going to be a very small fraction of the overall allocations compared to what GDK/GTK are doing anyway.

It means that there is a double pointer dereference (outer box to the trait object, trait object to the actual data), but also that shouldn't matter much in the greater scheme of things.

> 3. use static memory to store the closure (which isn't an issue since GDK stores the pointer it receives in static memory anyway)

I'm not sure what the threading guarantees of this function are but this might've needed a mutex for the global storage. You're right that it would work though as there can be at most one such callback at a time, and you get the destroy notify for clearing the static memory for it and replacing it with `NULL`.

It would be nicer to keep the general pattern of how we do this in other places though for consistency, i.e. 2. Otherwise 3. is also a valid option IMHO.

dhardy:
> I'm not sure what the threading guarantees of this function are

True. I guess it makes sense that the closure passed should be `Send` and that if we did use global memory, its usage should be protected.

There are quite a lot of events passed, but I guess the double-dereference thing is still not a big deal.

I pushed an updated version but belatedly realised it doesn't work (segfault).

dhardy:
Fixed. I didn't realise the type specification when boxing was important to generate the correct code!

Example usage:
```rust
        gdk::Event::set_handler(Some(|event: &mut gdk::Event| {
            println!("Event: {:?}", event);
            gtk::main_do_event(event);
        }));
```

src/event.rs
```diff
@@ -421,3 +445,20 @@ macro_rules! event_subtype {
         }
     }
 }
+
+extern "C" fn event_handler_trampoline(event: *mut ffi::GdkEvent, ptr: glib_ffi::gpointer) {
+    let mut event = unsafe { from_glib_none(event) };
```
sdroege: These functions should be `unsafe extern "C"` IMHO

src/event.rs
```diff
@@ -421,3 +445,20 @@ macro_rules! event_subtype {
...
+    if ptr != ptr::null_mut() {
+        let raw: *mut Box<dyn Fn(&mut Event) + Send +'static> = ptr as _;
+        let f: &(dyn Fn(&mut Event) + Send +'static) = unsafe { &**raw };
+        f(&mut event)
```
sdroege: Should the event be mutable? I don't know much about GDK events, can they be changed in a meaningful way and does that even work with the bindings like this?

src/event.rs
```diff
@@ -53,6 +54,29 @@ impl Event {
...
+            let ptr: *mut c_void = Box::into_raw(boxed) as *mut _;
+            unsafe{
+                ffi::gdk_event_handler_set(
+                    Some(event_handler_trampoline),
+                    ptr,
+                    Some(event_handler_destroy))
+            }
+        } else {
+            unsafe{
+                ffi::gdk_event_handler_set(None, ptr::null_mut(), None)
```
sdroege: How thread-safe is `gdk_event_handler_set()`? Can it be called from multiple threads at once safely, i.e. is it internally protected by a mutex? If not then this function still needs to be marked as `unsafe`.

src/event.rs
dhardy: I did this to match up with `gtk::main_do_event`. No idea really, though otherwise one must clone the event to call that. Maybe `main_do_event` shouldn't be mutable; would have to check GTK.

src/event.rs
sdroege: Here it would impossible to deadlock but that's not helping.

I assume it's only allowed to call from the main thread, so we could use the assertion here that is used all over GTK for making sure things are only called from the main thread. Unfortunately that part of the code is in the gtk crate, so not sure how to easily get that over here.

src/event.rs
sdroege: Should be investigated then :)

src/event.rs
sdroege: Yes but that marks the whole function unsafe. Everything in there is unsafe, not just the parts you previously put into an `unsafe` block :)

src/event.rs
```diff
@@ -53,6 +54,30 @@ impl Event {
...
+    /// 
+    /// The callback `handler` is called for each event. If `None`, event
+    /// handling is disabled.
+    /// 
+    /// This function is marked unsafe since it is a static function with no
+    /// protection against re-entry. It may be called by any thread but must
+    /// not be called concurrently with itself or with either `gtk::init` or
+    /// `gtk::Application::new`.
+    pub unsafe fn set_handler<F: Fn(&mut Event) + Send +'static>(handler: Option<F>) {
+        if let Some(handler) = handler {
```
EPashkin: Still need `assert_initialized_main_thread!();` in first line,
this also fix some thread safety problem.

dhardy: Not necessary; the GDK C code is simple, does only one read (from a null-initialised pointer) and is safe to call from any thread (though not reentrant):
```c
void 
gdk_event_handler_set (GdkEventFunc   func,
		       gpointer       data,
		       GDestroyNotify notify)
{
  if (_gdk_event_notify)
    (*_gdk_event_notify) (_gdk_event_data);

  _gdk_event_func = func;
  _gdk_event_data = data;
  _gdk_event_notify = notify;
}
```

EPashkin: Then your code don't pass CI 😉

EPashkin: Or you can use other macro, `skip_assert_initialized!();` if you sure about no problem in rust,
but IMHO better allow change only from main thread.

sdroege: Yes you can remove the `Send` requirement then.

I meant this one here specifically: https://github.com/gtk-rs/gtk/blob/b17cc04141daeb47971040a58a3d90af42d82c00/src/rt.rs#L59-L62
I don't think the GDK one is actually checking for the main thread

dhardy:
Okay, I added the `assert_initialized_main_thread` check and removed `unsafe` from `set_handler`.

The remaining question is about whether we should have a `&mut Event` or `&Event` in the callback.

One side of this is whether `gtk_main_do_event` actually mutates the event or not — this isn't simple to check, but since `rewrite_event_for_surface` actually clones the event before adjusting, it hints that the function doesn't mutate the event.

The other side is whether it even matters — is there any reason it *shouldn't* be `&mut Event`? By examining all calls to `_gdk_event_emit` it appears there is no technical issue since the event is always destroyed immediately after the handler is called.

Since it's more flexible I suggest we leave it as `&mut Event`?

EPashkin:
@dhardy About `&mut Event`: as have setters: `gdk_event_set_device` etc. seems we need mutable event.
We also have `pub fn get_scancode(&mut self) -> i32` and `pub fn get_pointer_emulated(&mut self) -> bool` but IMHO this just typo and better fixed (maybe not in this PR).

sdroege:
> The other side is whether it even matters — is there any reason it shouldn't be &mut Event? By examining all calls to _gdk_event_emit it appears there is no technical issue since the event is always destroyed immediately after the handler is called.
> 
> Since it's more flexible I suggest we leave it as &mut Event?

`&mut Event` means that it's also possible to replace the whole `Event` with a completely new event. Does that work?

EPashkin:
@dhardy Thanks.
IMHO this ready for merge now.

sdroege:
Yes it probably works

dhardy:
Great. Then I agree with @EPashkin that this is ready (unless you want me to squash).

Thanks for the help guys.

GuillaumeGomez:
No it's fine. Thanks a lot!

PR #239, opened 2018-09-04
Gen more things

GuillaumeGomez:
cc @sdroege @EPashkin


sdroege:
:+1:

EPashkin:
`gdk_property_change` has wrong Gdk-3.0.gir definition, so gir-files and sys need updated first,
or gir need fixed.
Current
```
        <parameter name="data" transfer-ownership="none">
          <doc xml:space="preserve">the data (a `guchar *`
  `gushort *`, or `gulong *`,
  depending on @format), cast to a `guchar *`.</doc>
          <type name="guint8" c:type="const guchar*"/>
        </parameter>

```
IMHO Type need be something like
```
          <array length="6" zero-terminated="0" c:type="const guint8*">
            <type name="guint8" c:type="guint8"/>
          </array>
```

GuillaumeGomez:
Updated. Bound the function manually.

EPashkin:
@GuillaumeGomez Thanks.

src/change_data.rs
```diff
@@ -0,0 +1,40 @@
...
+use libc;
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum ChangeData<'a> {
+    UChars(&'a [u8]),
+    UShorts(&'a [u16]),
+    ULongs(&'a [libc::c_ulong]),
+    UChar(u8),
+    UShort(u16),
+    ULong(libc::c_ulong),
```
EPashkin: Inner types don't need have `pub` ?

EPashkin: And here no `u32` cases

sdroege: u32/guint is not mentioned in the documentation of the function

GuillaumeGomez: > This seems to allow arbitrary types, I'm not sure how this function is supposed to work safely.

By limiting the number of entries. If we need more, we can always add them afterwards.

> Inner types don't need have `pub` ?

Woups... Thanks!

sdroege: Ok then. Do we know of any code actually using this binding? How do they use it?

GuillaumeGomez:
So? Is it ok for you?

sdroege:
Go for it :)

PR #246, opened 2018-09-16
FrameClock: fix segmentation fault after getting `FrameTimings`

fengalin:
## Use case and symptoms

In the `draw` callback of a `DrawingArea`, I want to retrieve the predicted presentation time for the frame being prepared:

``` rust
[...]
    let presentation_time = match da.get_frame_clock().unwrap().get_current_timings() {
        Some(frame_timings) => {
            let time = frame_timings.get_predicted_presentation_time() as u64;
            // std::mem::forget(frame_timings); // Uncomment to avoid the segmentation fault
            time
        }
        None => {
            debug!("can't get frame timings");
            return Inhibit(false);
        }
    };
[...]
```

With the above code, I get a couple of frames ok, and then a segmentation fault with a critical GDK message:

```
(.:2995): Gdk-CRITICAL **: 12:01:03.723: gdk_frame_timings_unref: assertion 'timings->ref_count > 0' failed

Thread 1 "media-toc" received signal SIGSEGV, Segmentation fault.
```

Uncommenting `std::mem::forget(frame_timings)` prevents the segmentation fault from happening.

## Analysis

[`gdk_frame_clock_get_current_timings`](https://github.com/GNOME/gtk/blob/8fd2d461fce6da655060880fdb183a9f1e0b957c/gdk/gdkframeclock.c#L485) makes use of [`gdk_frame_clock_get_timings`](https://github.com/GNOME/gtk/blob/8fd2d461fce6da655060880fdb183a9f1e0b957c/gdk/gdkframeclock.c#L451), which [returns a pointer directly from the `priv->timings` array](https://github.com/GNOME/gtk/blob/8fd2d461fce6da655060880fdb183a9f1e0b957c/gdk/gdkframeclock.c#L469), without increasing the reference counter on the returned `GdkFrameTimings`.

Since the documentation doesn't explicitly sates that the user should `unref` the returned pointer after use, I guess this is intentional.

My understanding is that in the Rust GDK binding, [`FrameTimings` is treated as a reference counted object](https://github.com/gtk-rs/gdk/blob/42e869730228b29ef07f95883f34ed9e5a2daa01/src/auto/frame_timings.rs#L14), which it is, but which causes Rust to decrement the reference counter when the Rust object is dropped, which deallocates the C object in the `priv->timings` array, causing the segmentation fault when the C code attempts to access it later.

## Proposal

I'm still far from being fluent with GLib and its binding. My best proposal so far is to borrow the returned object. The C code seems to be the same in [current master](https://github.com/GNOME/gtk/blob/d7a5dcba0b06a51781b4c7207f47f2155646dbc0/gdk/gdkframeclock.c#L448). Should it change, we could add a feature gate based on the targetted GDK (GTK) version.


GuillaumeGomez:
I agree with your analysis and your fix. So unless anyone else has a better proposal, I think we'll go for this one. Thanks!

cc @sdroege @EPashkin

src/frame_clock.rs
```diff
@@ -16,4 +18,18 @@ impl FrameClock {
             (refresh_interval, presentation_time)
         }
     }
+
+    #[cfg(any(feature = "v3_8", feature = "dox"))]
+    pub fn get_current_timings(&self) -> Option<FrameTimings> {
+        unsafe {
+            from_glib_borrow(ffi::gdk_frame_clock_get_current_timings(self.to_glib_none().0))
```
sdroege: This is wrong. You must never ever return a `from_glib_borrow()` return value to safe code. It's borrowed but without making sure the owner is kept alive.

This needs to be `from_glib_none()`. Which is also what the C documentation seems to suggest, not sure why the `.gir` file says `transfer full`.

EPashkin: If `from_glib_none` works then change better be done in https://github.com/gtk-rs/gir-files with fix.sh instead

EPashkin: With something like
```
# fix wrong "full" transfer ownership
xmlstarlet ed -P -L \
	-u '//_:method[@c:identifier="gdk_frame_clock_get_current_timings"]/_:return-value/@transfer-ownership' -v "none" \
	-u '//_:method[@c:identifier="gdk_frame_clock_get_timings"]/_:return-value/@transfer-ownership' -v "none" \
	Gdk-3.0.gir
```

Issue #250, opened 2018-10-13
Proposal: use an `Option` for return value of `get_[predicted_]presentation_time()`


fengalin:
`get_presentation_time` & `get_predicted_presentation_time` [can both return `0` when the value is not available](https://developer.gnome.org/gdk3/stable/gdk3-GdkFrameTimings.html#gdk-frame-timings-get-predicted-presentation-time).

In Rust, returning `None` in this case would make it clear that the value shouldn't be used as is. I tried changing the signature to return an `Option` using Gir's `nullable = true`, but it didn't work.

What do you think about this change? If you agree, I'll open a PR.
Is there a Gir-way to handle this or should I use a manual implementation?

EPashkin:
IMHO sound reasonable.
Can be done only with manual implementation.

fengalin:
@EPashkin ok thanks. I'll do this soon then.

fengalin:
Fixed per #251

PR #251, opened 2018-10-14
FrameTimings: use an `Option` when returning presentation time

fengalin:
`get_presentation_time` and `get_predicted_presentation_time` can both return `0` when the returned value is not available. In Rust, using an `Option` and returning `None` would make it clear that the value shouldn't be used as is.

See #250

### DIFF
--- a/src/frame_timings.rs
+++ b/src/frame_timings.rs
@@ -1,0 +1,32 @@
+// Copyright 2018, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use glib::translate::*;
+use ffi;
+use std::num::NonZeroU64;
+use FrameTimings;
+
+impl FrameTimings {
+    #[cfg(any(feature = "v3_8", feature = "dox"))]
+    pub fn get_predicted_presentation_time(&self) -> Option<NonZeroU64> {
+        let predicted_presentation_time = unsafe {
+            ffi::gdk_frame_timings_get_predicted_presentation_time(self.to_glib_none().0)
+        };
+        // assuming presentation time is always positive
+        assert!(predicted_presentation_time >= 0);
+        // `0` means the value is not available
+        NonZeroU64::new(predicted_presentation_time as u64)
+    }
+
+    #[cfg(any(feature = "v3_8", feature = "dox"))]
+    pub fn get_presentation_time(&self) -> Option<NonZeroU64> {
+        let presentation_time = unsafe {
+            ffi::gdk_frame_timings_get_presentation_time(self.to_glib_none().0)
+        };
+        // assuming presentation time is always positive
+        assert!(presentation_time >= 0);
+        // `0` means the value is not available
+        NonZeroU64::new(presentation_time as u64)
+    }
+}
--- a/Gir.toml
+++ b/Gir.toml
@@ -286,6 +286,14 @@ status = "generate"
 name = "Gdk.FrameTimings"
 status = "generate"
 version = "3.8"
+    [[object.function]]
+    name = "get_predicted_presentation_time"
+    # Use an `Option` for the return value
+    ignore = true
+    [[object.function]]
+    name = "get_presentation_time"
+    # Use an `Option` for the return value
+    ignore = true
 
 [[object]]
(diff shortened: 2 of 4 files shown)


EPashkin:
Please update gir submodule to or do regen with gir c385982dbed5fdbdf1ae6f95c885a475cafc93f6

EPashkin:
Other that this nit all LFGM

fengalin:
Oh! Sure, ' will do that.

fengalin:
Fixed, sorry.

EPashkin:
@fengalin Thanks.

EPashkin:
@GuillaumeGomez 👍 
cc @sdroege

sdroege:
Should probably be an `Option<NonZero<_>>` then?

sdroege:
Are negative values valid?

fengalin:
> Should probably be an Option<NonZero<_>> then?

I'm not sure about this. I agree this would be a nice addition, but searching for [`NonZero` in the `std`](https://doc.rust-lang.org/std/index.html?search=NonZero), I could only spot usigned integers candidates.

> Are negative values valid?

It's not clear to me whether the timings might be negative. However, the [GDK code](https://github.com/GNOME/gtk/blob/master/gdk/gdkframeclock.c) seems to always make use of `if (presentation_time != 0)` when handling similar cases, so I wouldn't take the risk of assuming they are exclusively positive.

GuillaumeGomez:
A question on this: should we add a crate this small as a dependency or should we just write the needed type and go for it? We already have quite a lot of dependencies after all and the rendering of the docs won't appreciate much (I just keep "top-level" crates).

What do you think of this?

sdroege:
I'd just depend on it, that's what crates are for. If rustdoc has problems with many dependencies then that's a bug in rustdoc that has to be fixed there.

GuillaumeGomez:
Not a bug in rustdoc, a "bug" in our documentation generation.

sdroege:
What exactly is the problem?

GuillaumeGomez:
Not really a problem, just a nuisance. I'll just have to update the release script, but from now on, we will have to add the crate to gtk-rs docs.

sdroege:
:+1:

sdroege:
I still think it's an API bug that this is not a (unsigned!) `guint64` on the C side. Negative values make no sense really, or 0 would make equally much sense as a valid value :)

fengalin:
I just realized I didn't conditioned the dependency on `NoneZeroI64` to the proper version... Doing it now.

fengalin:
> I still think it's an API bug that this is not a (unsigned!) guint64 on the C side. Negative values make no sense really, or 0 would make equally much sense as a valid value :)

My guess is that it returns a signed in order to ease arithmetic operations, but I agree timings shouldn't be negative.

Do you think we should fix this in the Rust binding?

sdroege:
We could, yes. Casting is annoying in Rust and implicit in C

fengalin:
All right, I kept the commit stack for the record. I can squash if everyone agrees with the unsigned approach.

src/frame_timings.rs
```diff
@@ -4,21 +4,21 @@
...
 use FrameTimings;
 
 impl FrameTimings {
     #[cfg(any(feature = "v3_8", feature = "dox"))]
-    pub fn get_predicted_presentation_time(&self) -> Option<NonZeroI64> {
-        NonZeroI64::new(unsafe {
-            ffi::gdk_frame_timings_get_predicted_presentation_time(self.to_glib_none().0)
+    pub fn get_predicted_presentation_time(&self) -> Option<NonZeroU64> {
+        NonZeroU64::new(unsafe {
+            ffi::gdk_frame_timings_get_predicted_presentation_time(self.to_glib_none().0) as u64
```
sdroege: Maybe add an assertion here for negative numbers, and below

fengalin: Of course!

fengalin:
@EPashkin & @GuillaumeGomez, are you ok with latest version?

EPashkin:
@fengalin Yes, Thanks

GuillaumeGomez:
Yep, all good for me, thanks! Ping me once squash is done and then I merge!

fengalin:
@GuillaumeGomez: squashed and all green now! :)

GuillaumeGomez:
Thanks!

PR #252, opened 2018-10-17
FrameTimings: use an `Option` when returning refresh_interval

fengalin:
`gdk_frame_timings_get_refresh_interval` [returns `0` when the value is not available](https://developer.gnome.org/gdk3/stable/gdk3-GdkFrameTimings.html#gdk-frame-timings-get-refresh-interval).

Same as https://github.com/gtk-rs/gdk/pull/251.

### DIFF
--- a/src/frame_timings.rs
+++ b/src/frame_timings.rs
@@ -29,4 +29,15 @@ impl FrameTimings {
         // `0` means the value is not available
         NonZeroU64::new(presentation_time as u64)
     }
+
+    #[cfg(any(feature = "v3_8", feature = "dox"))]
+    pub fn get_refresh_interval(&self) -> Option<NonZeroU64> {
+        let refresh_interval = unsafe {
+            ffi::gdk_frame_timings_get_refresh_interval(self.to_glib_none().0)
+        };
+        // assuming refresh interval is always positive
+        assert!(refresh_interval >= 0);
+        // `0` means the value is not available
+        NonZeroU64::new(refresh_interval as u64)
+    }
 }
--- a/Gir.toml
+++ b/Gir.toml
@@ -294,6 +294,10 @@ version = "3.8"
     name = "get_presentation_time"
     # Use an `Option` for the return value
     ignore = true
+    [[object.function]]
+    name = "get_refresh_interval"
+    # Use an `Option` for the return value
+    ignore = true
 
 [[object]]
 name = "Gdk.Monitor"
--- a/src/auto/frame_timings.rs
+++ b/src/auto/frame_timings.rs
@@ -41,11 +41,4 @@ impl FrameTimings {
             ffi::gdk_frame_timings_get_frame_time(self.to_glib_none().0)
         }
     }
-
-    #[cfg(any(feature = "v3_8", feature = "dox"))]
-    pub fn get_refresh_interval(&self) -> i64 {
-        unsafe {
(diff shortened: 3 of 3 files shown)


EPashkin:
@fengalin You seems right again, 👍

fengalin:
I wish I had seen this as part of #251, but better later than never!

src/frame_timings.rs
```diff
@@ -29,4 +29,15 @@ impl FrameTimings {
...
         NonZeroU64::new(presentation_time as u64)
     }
+
+    #[cfg(any(feature = "v3_8", feature = "dox"))]
+    pub fn get_refresh_interval(&self) -> Option<NonZeroU64> {
+        let refresh_interval = unsafe {
+            ffi::gdk_frame_timings_get_refresh_interval(self.to_glib_none().0)
+        };
+        // assuming refresh interval is always positive
+        assert!(refresh_interval >= 0);
```
GuillaumeGomez: If not positive, shouldn't it return `None` instead?

fengalin: As you prefer. I followed the observation in the second comment [there](https://github.com/gtk-rs/gdk/pull/251#issuecomment-429768311) and the subsequent conversation, especially [this comment](https://github.com/gtk-rs/gdk/pull/251#issuecomment-429925655).

GuillaumeGomez: I just prefer `panic` calls as much as possible.

fengalin: In that case, that should also be the behaviour for #251. My point is that according to the doc and the code I could read, only `0` is considered as "the value is not available". I agree a timestamp could be negative, but an interval which is computed from the succession of frame presentation times shouldn't.

@sdroege & @EPashkin, what do you think?

GuillaumeGomez: Oh indeed, I didn't pay attention to that!

sdroege: IMHO this should panic as it's unexpected behaviour for the bindings and `None` has a different meaning. A negative value would have to be returned somehow if it is valid, but from our understanding it is never valid.

If this assumption is ever wrong, it has to fail hard and we have to fix it at the bindings level.

The way how the code is right now seems correct to me.

GuillaumeGomez: Uncatchable errors seems like a really bad issue for me. `None` and `Err` variants have been created for cases like this... We can always return `None` and use `eprintln!` to show that something problematic happened like GTK does.

sdroege: This is an uncatchable error by definition. It's like a function returning `NULL` when it really shouldn't and we don't expect it to either.

Making users suffer by returning an enum for all 3 cases does not seem like a good idea here. This case should just never happen at all, users are not supposed to handle it because it's a problem on our side and not on their side.

fengalin: I believe this is not an error in the sense "something the user should handle", but rather a bug either on the GTK side or due to our interpretation. Just like @sdroege said, if we were wrong, we should fix it. In that sense, the assertion violation message is _a propos_ WRT this context.

My tests are far from being enough to serve as a proof that this can't happen, but at least it didn't show up with a couple of videos on a Linux host and a Windows host.

Still, I have no hard feelings with this and could just as well use a `Result` and `eprintln!` in case the value is negative. If you think that should be done, I'll also open another PR to fix #251.

GuillaumeGomez: No, you both convinced me. I'll go for this code then!

PR #259, opened 2019-01-13
Support const array of gdk::Atom

EPashkin:
Will be needed for https://github.com/gtk-rs/gir-files/pull/32, but can be done before.

cc @GuillaumeGomez , @sdroege

### DIFF
--- a/src/atom.rs
+++ b/src/atom.rs
@@ -94,7 +94,7 @@ impl<'a> ToGlibContainerFromSlice<'a, *mut ffi::GdkAtom> for &'a Atom {
         let v: Vec<_> = t.iter().map(|s| s.to_glib_none()).collect();
 
         let v_ptr = unsafe {
-            let v_ptr = glib_ffi::g_malloc0(mem::size_of::<ffi::GdkAtom>() * t.len() + 1) as *mut ffi::GdkAtom;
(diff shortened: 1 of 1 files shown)


src/atom.rs
```diff
@@ -113,6 +113,44 @@ impl<'a> ToGlibContainerFromSlice<'a, *mut ffi::GdkAtom> for &'a Atom {
...
+        (v_ptr.as_ptr() as *const ffi::GdkAtom, (v, Some(v_ptr)))
+    }
+
+    fn to_glib_container_from_slice(t: &'a [&'a Atom]) -> (*const ffi::GdkAtom, Self::Storage) {
+        skip_assert_initialized!();
+
+        let v: Vec<_> = t.iter().map(|s| s.to_glib_none()).collect();
+
+        let v_ptr = unsafe {
+            let v_ptr = glib_ffi::g_malloc0(mem::size_of::<ffi::GdkAtom>() * t.len() + 1) as *mut ffi::GdkAtom;
```
GuillaumeGomez: I think the correct syntax here would be (wrapping the `t.len() + 1`):

```
let v_ptr = glib_ffi::g_malloc0(mem::size_of::<ffi::GdkAtom>() * (t.len() + 1)) as *mut ffi::GdkAtom;
```

EPashkin: Good catch, thanks

EPashkin: Fixed

GuillaumeGomez:
Thanks!

PR #266, opened 2019-01-28
Don't box closures twice

sdroege:


EPashkin:
👍

sdroege:
What does that mean? :)

sdroege:
> I don't know, probably old check_init_asserts(gtks check different names), or it bad with closures.

Ok but this is nothing new from my PR so let's worry about that another time :)

EPashkin:
Agree.
We fails on new `event_handler_destroy` and `event_handler_trampoline` and want assert macro here.
Seems constraint broke regex

sdroege:
@GuillaumeGomez good to go :)

PR #282, opened 2019-02-24
Regen

GuillaumeGomez:
cc @EPashkin @sdroege


src/cairo_interaction.rs
```diff
@@ -68,9 +68,8 @@ impl ContextExt for Context {
         }
     }
 
-    fn cairo_surface_create_from_pixbuf<'a, P: Into<Option<&'a Window>>>(pixbuf: &Pixbuf, scale: i32, for_window: P) -> Option<Surface> {
+    fn cairo_surface_create_from_pixbuf(pixbuf: &Pixbuf, scale: i32, for_window: Option<&Window>) -> Option<Surface> {
```
EPashkin: @sdroege Not related to this PR, you don't know why `gdk::Window` don't used with `IsA` ?

EPashkin:
👍

src/cairo_interaction.rs
GuillaumeGomez: That's a good question actually.

EPashkin: Oh, this not generated code, so it related to this PR actually 😉

EPashkin:
@GuillaumeGomez not better add PR to examples first?

GuillaumeGomez:
@EPashkin I'll do it next.

src/cairo_interaction.rs
sdroege: Should be `IsA`

src/auto/device.rs
```diff
@@ -206,8 +206,7 @@ impl Device {
     }
 
     #[cfg_attr(feature = "v3_20", deprecated)]
-    pub fn grab<'a, P: IsA<Window>, Q: Into<Option<&'a Cursor>>>(&self, window: &P, grab_ownership: GrabOwnership, owner_events: bool, event_mask: EventMask, cursor: Q, time_: u32) -> GrabStatus {
-        let cursor = cursor.into();
+    pub fn grab<P: IsA<Window>>(&self, window: &P, grab_ownership: GrabOwnership, owner_events: bool, event_mask: EventMask, cursor: Option<&Cursor>, time_: u32) -> GrabStatus {
```
sdroege: Should be `IsA<Cursor>` or not?

src/event.rs
```diff
@@ -248,9 +248,9 @@ impl Event {
         }
     }
 
-    pub fn set_screen<'a, T: Into<Option<&'a Screen>>>(&mut self, screen: T) {
+    pub fn set_screen(&mut self, screen: Option<&Screen>) {
```
sdroege: These should also all be `IsA<_>` or not?

sdroege:
Looks good apart from the missing `IsA<_>`

src/event.rs
EPashkin: Seems it final
https://github.com/gtk-rs/gdk/blob/47bdccab08a8366788b74e2ab1bc97f6f7269d81/src/auto/app_launch_context.rs#L46

sdroege: That's good

sdroege: Should double-check for all these types then :)

src/auto/device.rs
GuillaumeGomez: That's automatic generation so if problem there is, it lies in gir.

EPashkin:
Only maybe non-final type used in manual code is `FrameClock`,
not sure that it need `FrameClockExtManual` or `final_type = true`.

GuillaumeGomez:
I'll merge this one then meanwhile. Waiting for the next releases. :)

PR #285, opened 2019-03-18
Fields have been generated so we can uncommented those methods

GuillaumeGomez:
cc @sdroege @EPashkin

### DIFF
--- a/src/event_key.rs
+++ b/src/event_key.rs
@@ -38,8 +38,7 @@ impl EventKey {
         self.as_ref().group
     }
 
-    // TODO: to be uncommented when the field will be generated
-    // pub fn get_is_modifier(&self) -> bool {
-    //     from_glib(self.as_ref().is_modifier)
-    // }
+    pub fn get_is_modifier(&self) -> bool {
+        self.as_ref().is_modifier & 1 != 0
+    }
 }
--- a/src/event_scroll.rs
+++ b/src/event_scroll.rs
@@ -46,8 +46,7 @@ impl EventScroll {
         (dx, dy)
     }
 
-    // TODO: to be uncommented when the field will be generated
-    // pub fn get_is_stop(&self) -> bool {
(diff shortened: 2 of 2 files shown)


src/event_key.rs
```diff
@@ -38,8 +39,7 @@ impl EventKey {
         self.as_ref().group
     }
 
-    // TODO: to be uncommented when the field will be generated
-    // pub fn get_is_modifier(&self) -> bool {
-    //     from_glib(self.as_ref().is_modifier)
-    // }
+    pub fn get_is_modifier(&self) -> bool {
+        self.as_ref().is_modifier != 0
```
EPashkin: Its bitfield, so IMHO better something like `self.as_ref().is_modifier & 1 != 0`,
same for `is_stop`

GuillaumeGomez:
Updated.

src/auto/versions.txt
```diff
@@ -1,2 +1,2 @@
-Generated by gir (https://github.com/gtk-rs/gir @ 914ca73)
+Generated by gir (https://github.com/gtk-rs/gir @ 1ae7210)
```
EPashkin: This file changed but submodule not updated.

src/event_key.rs
```diff
@@ -4,6 +4,7 @@
 
 use gdk_sys;
 use glib::translate::*;
+use libc;
```
EPashkin: IMHO this use not used.

EPashkin: https://travis-ci.org/gtk-rs/gdk/jobs/508079289#L673

GuillaumeGomez:
Updated.

EPashkin:
Thanks, 👍

src/event_key.rs
GuillaumeGomez: No it wasn't. Weird that I added it... Good catch!

sdroege:
:+1:

PR #286, opened 2019-03-23
add doc rule

GuillaumeGomez:


EPashkin:
👍 and for all next

GuillaumeGomez:
Thanks! I'll still wait for CIs to avoid having failures everywhere (even though we'll still have...).

PR #289, opened 2019-04-19
Fix travis build

GuillaumeGomez:


EPashkin:
Good, one travis fully fine with only PKG_CONFIG_PATH

EPashkin:
@GuillaumeGomez CI passed here, 👍 for this and all with only PKG_CONFIG_PATH

PR #290, opened 2019-05-10
Move functions into traits when needed

GuillaumeGomez:
cc @sdroege @EPashkin

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -4,11 +4,16 @@
 
 use gdk_sys;
 use glib::translate::*;
+use glib::object::IsA;
 use AxisUse;
 use Device;
 
-impl Device {
-    pub fn get_axis(&self, axes: &mut [f64], use_: AxisUse, value: &mut f64) -> bool {
-        unsafe { from_glib(gdk_sys::gdk_device_get_axis(self.to_glib_none().0, axes.as_mut_ptr(), use_.to_glib(), value)) }
+pub trait DeviceExtManual: 'static {
+    fn get_axis(&self, axes: &mut [f64], use_: AxisUse, value: &mut f64) -> bool;
+}
+
+impl<O: IsA<Device>> DeviceExtManual for O {
+    fn get_axis(&self, axes: &mut [f64], use_: AxisUse, value: &mut f64) -> bool {
+        unsafe { from_glib(gdk_sys::gdk_device_get_axis(self.as_ref().to_glib_none().0, axes.as_mut_ptr(), use_.to_glib(), value)) }
     }
 }
--- a/src/frame_clock.rs
+++ b/src/frame_clock.rs
@@ -5,13 +5,18 @@
 use gdk_sys;
 use glib::translate::*;
 use FrameClock;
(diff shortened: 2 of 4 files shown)


src/device.rs
```diff
@@ -4,11 +4,16 @@
...
 use Device;
 
-impl Device {
-    pub fn get_axis(&self, axes: &mut [f64], use_: AxisUse, value: &mut f64) -> bool {
-        unsafe { from_glib(gdk_sys::gdk_device_get_axis(self.to_glib_none().0, axes.as_mut_ptr(), use_.to_glib(), value)) }
+pub trait DeviceExtManual: 'static {
+    fn get_axis(&self, axes: &mut [f64], use_: AxisUse, value: &mut f64) -> bool;
+}
+
+impl<O: IsA<Device>> DeviceExtManual for O {
```
sdroege: `gdk::Device` is not subclassable outside of `gdk` and there are no subclasses inside `gdk`, are there?

src/frame_clock.rs
```diff
@@ -5,13 +5,18 @@
 use gdk_sys;
 use glib::translate::*;
 use FrameClock;
+use glib::object::IsA;
 
-impl FrameClock {
-    pub fn get_refresh_info(&self, base_time: i64) -> (i64, i64) {
+pub trait FrameClockExtManual: 'static {
```
sdroege: Same

EPashkin: Then we need remove trait from generated code too

sdroege: Interesting, why is it generated at all? It shouldn't because the configuration is not overridden, and the default behaviour is to only generate the trait if there is a class struct (-> it is possible to create subclasses outside this crate) or there are known subclasses (-> there are subclasses inside this crate).

EPashkin: There class for FrameClock, but this class is disguised and `mark_final_types` don't checks that
```<record name="FrameClockClass" c:type="GdkFrameClockClass" disguised="1" glib:is-gtype-struct-for="FrameClock">```

Comment (sdroege) [reply src/frame_clock.rs]: Ah we need to fix that in gir then

Comment (EPashkin) [reply src/frame_clock.rs]: I tried fix that, but found this comment in parser.
Still think that to do.
```
        // Currently g-ir-scanner is too eager to mark all typedef to non-complete
        // types as disguised, so we limit this behaviour to the two most common cases.
        // https://gitlab.gnome.org/GNOME/gobject-introspection/issues/101
        let disguised = c_type == "GdkAtom" || c_type == "GIConv";
```

sdroege: I meant only ignoring the disguised ones for the class detection code

EPashkin: Problem that this is place in parser where disguised  must be parsed.
But problem seems solved and we can use disguised="1".
I will add issue to collect all PRs

EPashkin: https://github.com/gtk-rs/gir/issues/760

GuillaumeGomez:
Can be closed then!

PR #293, opened 2019-06-08
Remove traits for disguised types without children

EPashkin:
Part of gtk-rs/gir#760

cc @GuillaumeGomez


GuillaumeGomez:
Thanks a lot! Sorry for the mess...

PR #299, opened 2019-06-23
Fix docs for manual functions [ci skip]

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/797

cc @GuillaumeGomez

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -185,16 +185,6 @@ impl<O: IsA<Window>> WindowExtManual for O {
         &mut *(pointer as *mut T)
     }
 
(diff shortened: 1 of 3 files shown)


src/window.rs
```diff
@@ -185,16 +185,6 @@ impl<O: IsA<Window>> WindowExtManual for O {
         &mut *(pointer as *mut T)
     }
 
-    fn set_geometry_hints(&self, geometry: &gdk_sys::GdkGeometry, geom_mask: WindowHints) {
```
EPashkin: This is generated as we have manual Geometry

GuillaumeGomez:
Thanks!

Issue #308, opened 2019-08-12
Clean up ciaro interactions


bilelmoussaoui:
Currently, some methods are implemented as part of the ContextExt which shouldn't be the case.

The PixbufExt isn't loaded on prelude either.

sdroege:
* `cairo_draw_from_gl()` should take the context as `&self`
* `cairo_surface_create_from_pixbuf()` should be removed from `ContextExt` (it's in `PixbufExt` where it should be)
* The traits should get `Gdk` as prefix of their name to prevent conflicts

bilelmoussaoui:
Thanks! Will get a MR for this later today

sdroege:
> Thanks! Will get a MR for this later today

Long day :) I've created https://github.com/gtk-rs/gdk/pull/312 for this now, please review if you have some time

bilelmoussaoui:
Thanks for taking the time to fix it. Completely forgot about it

PR #309, opened 2019-08-12
Cairo interactions: auto load Pixbuf & Surface Exts

bilelmoussaoui:
Part 1 of #308

### DIFF
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -5,5 +5,5 @@
 //! Traits intended for blanket imports.
 
 pub use auto::traits::*;
-pub use cairo_interaction::ContextExt;
+pub use cairo_interaction::{ContextExt, PixbufExt, SurfaceExt};
 pub use window::WindowExtManual;


GuillaumeGomez:
I don't understand the need for these changes, even after reading the issue...

bilelmoussaoui:
Because its currently impossible to use those functions. Some need to be moved out Of the ContextExt which I want to send as a different PR

GuillaumeGomez:
Even if you import the traits in your code? Can you show us an example with the problematic code please?

bilelmoussaoui:
Without this patch I can't draw on a surface and restore the latest surface which is used on drawing user selected rectangle see https://gitlab.gnome.org/bilelmoussaoui/censor/blob/df8807078d50daf9aa3c2a631c0a4c1515a6d914/src/widgets/drawer.rs#L161 the code is a bit messy for now

sdroege:
The change is correct, the traits must all be exported from the prelude like all other traits.

GuillaumeGomez:
Makes sense indeed. Thanks @bilelmoussaoui !

sdroege:
@GuillaumeGomez or @bilelmoussaoui can you create an issue here https://github.com/gtk-rs/gtk4/issues or a PR for doing the same change?

bilelmoussaoui:
Will do.

PR #312, opened 2019-09-05
Fix up cairo interaction extension traits

sdroege:
- Remove duplicated ContextExt::cairo_surface_create_from_pixbuf(). It
  already exists as PixbufExt::create_surface()
- Rename ContextExt::cairo_draw_from_gl() to draw_from_gl() and take the
  context as &self
- Mark ContextExt::draw_from_gl as unsafe! It takes random integers for
  GL resources, which are basically like pointers and must be taken care
  of to be correct and valid.
- Change all functions taking a gdk::Window to take a IsA<gdk::Window>
- Rename all traits to include the Gdk prefix for preventing naming
  conflicts.

Fixes https://github.com/gtk-rs/gdk/issues/308

### DIFF
--- a/src/cairo_interaction.rs
+++ b/src/cairo_interaction.rs
@@ -10,11 +10,11 @@ use glib::object::IsA;
 use glib::translate::*;
 use {Rectangle, Window, RGBA};
 
-pub trait SurfaceExt {
+pub trait GdkSurfaceExt {
     fn create_region(&self) -> Option<Region>;
 }
 
-impl SurfaceExt for Surface {
+impl GdkSurfaceExt for Surface {
     fn create_region(&self) -> Option<Region> {
         unsafe {
             from_glib_full(gdk_sys::gdk_cairo_region_create_from_surface(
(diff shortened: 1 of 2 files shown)


src/cairo_interaction.rs
```diff
@@ -10,11 +10,11 @@ use glib::object::IsA;
 use glib::translate::*;
 use {Rectangle, Window, RGBA};
 
-pub trait SurfaceExt {
+pub trait GdkSurfaceExt {
```
GuillaumeGomez: "Gdk"SurfaceExt?

GuillaumeGomez: It comes from GDK but do we really want to name it like this?

EPashkin: IMHO yes, as  "it comes from GDK"

sdroege: Yes?

EPashkin:
@sdroege Thanks, 👍

src/cairo_interaction.rs
GuillaumeGomez: Fine then

sdroege: Just like we have GtkWindowExt for example

GuillaumeGomez: This is all too complicated for me.

sdroege:
@GuillaumeGomez all green

GuillaumeGomez:
Thanks!

PR #326, opened 2020-01-01
Fix osx CI build

GuillaumeGomez:
Let's see if it fixes the CI issue...


sdroege:
I don't understand this PR. What does it fix and why does removing Python 2 fix something?

GuillaumeGomez:
Because brew is not good for stuff like that I assume...

Issue #304, opened 2019-07-12
threads_add_idle asserts it is running on the main thread


bmteller:
I assume it is like this because the auto-generating of the bindings. But I think this function is safe to call when not on the main thread.

cmyr:
I also just ran into this.

Not only is this function safe to call from other threads, it is the recommended way of communicating back to the main thread from another thread.

I'd be happy to come up with a patch, but I'm not sure how that works with the generated bindings?

EPashkin:
@cmyr: So issue need be renamed to "threads_add_idle asserts **if** is running **not** on the main thread" ?

EPashkin:
It need fix in gir, I will do it on the weekend

sdroege:
@cmyr You can use [`glib::source::idle_add()`](https://gtk-rs.org/docs/glib/source/fn.idle_add.html), or for a slightly different pattern [`glib::MainContext::channel()`](https://gtk-rs.org/docs/glib/struct.MainContext.html#method.channel) or any approach using futures (the main context is a futures executor).

I think we should simply ignore this and similar functions from gdk. They don't bring any additional advantage.

sdroege:
> I think we should simply ignore this and similar functions from gdk. They don't bring any additional advantage.

I.e. all `gdk_threads_.*` functions should be ignored.

sdroege:
@EPashkin You make a PR for that?

cmyr:
`idle_add()` looks like it would probably work. I don't have enough context to offer a strong opinion; I don't use gtk much, I'm just maintaining some code.

In terms of naming of this thread title, well lets not spend too much time on it 😉 but I think the current phrasing is reasonable; there is an assert that checks if the function is running on the main thread, which fails otherwise, even though it should be okay to call this function on other threads.

sdroege:
Yeah, we're just going to remove the functions :) They're only duplicating the behaviour of the functions with the same names from glib.

PR #333, opened 2020-02-13
Ignore thread_add and threads_add_timeout in favor of glibs analog

EPashkin:
Part of https://github.com/gtk-rs/gdk/issues/304

@sdroege Maybe instead ignoring we need fixing assert: https://developer.gnome.org/gdk3/stable/gdk3-Threads.html#gdk-threads-add-idle-full

cc @GuillaumeGomez


sdroege:
> [](https://github.com/gtk-rs/gdk/issues/304)
> 
> @sdroege Maybe instead ignoring we need fixing assert: https://developer.gnome.org/gdk3/stable/gdk3-Threads.html#gdk-threads-add-idle-full

All GTK/GDK things are not thread-safe anyway and the example there is rather contrived. You can't do what is described there with safe Rust code with the bindings, and it's also unsafe to do in C on Windows and macOS at least.

EPashkin:
@sdroege Thanks for the clarification

sdroege:
Let's get this in then? @GuillaumeGomez ?

GuillaumeGomez:
Thanks @EPashkin !

PR #341, opened 2020-04-10
Regen

GuillaumeGomez:
Part of gtk-rs/gir#896.

cc @EPashkin @sdroege


src/keys.rs
```diff
@@ -16,11 +15,6 @@ pub fn keyval_name(keyval: u32) -> Option<GString> {
 pub fn keyval_to_unicode(keyval: u32) -> Option<char> {
     skip_assert_initialized!();
     unsafe {
-        let c: char = transmute(gdk_sys::gdk_keyval_to_unicode(keyval));
-        if c != '\0' {
-            Some(c)
-        } else {
-            None
-        }
+        ::std::char::from_u32(gdk_sys::gdk_keyval_to_unicode(keyval))
```
GuillaumeGomez, line 18: Should I still check for `'\0'` in here too?

sdroege: Yes because that's what the function documentation says. It returns 0 if it didn't find a corresponding unicode character

PR #342, opened 2020-04-10
window: store event_mask using EventMask

### LINKED ISSUE #338: gdk::WindowAttr should contain an EventMask in the event_mask field
Currently the `event_mask` field has type `i32`. The `EventMask` struct exists and can be built as a bitfield, and then converted using `.bits()`, yielding an `u32` which then needs to be cast in order to be used in the `event_mask` field. This is somewhat unwieldy, and it would likely be preferable to just have the `WindowAttr` struct contain the `EventMask`.

gdesmott:
Fix #338

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -10,6 +10,7 @@ use glib::translate::*;
 use libc::{c_char, c_int};
 use std::ptr;
 use Cursor;
+use EventMask;
 use Visual;
 use Window;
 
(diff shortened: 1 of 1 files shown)


src/window.rs
```diff
@@ -94,7 +96,7 @@ impl<'a> ToGlibPtr<'a, *mut gdk_sys::GdkWindowAttr> for WindowAttr {
 
         let mut attrs = Box::new(gdk_sys::GdkWindowAttr {
             title: title.0 as *mut c_char,
-            event_mask: self.event_mask,
+            event_mask: i32::try_from(self.event_mask.bits()).unwrap(),
```
sdroege, line 99: This would fail if the top bit is set. Should just be a cast

gdesmott: fixed.

EPashkin:
👍

sdroege:
@GuillaumeGomez Green CI :)

GuillaumeGomez:
Thanks @gdesmott !

Issue #345, opened 2020-04-27
Nicer bindings for Key(val) related functions


afranke:
Key has no method and we end up with e.g. `keyval_to_unicode(ev.get_keyval())` instead of `ev.get_keyval().to_unicode()`.

GuillaumeGomez:
This is a good point, thanks for bringing it up!

PR #346, opened 2020-04-30
Nicer bindings for Key(val) related functions

danigm:
Fix https://github.com/gtk-rs/gdk/issues/345


tests/check_event.rs
```diff
@@ -0,0 +1,25 @@
...
+    assert_eq!(keyval, Some('A'));
+    assert_eq!(keyval, keyval2);
+
+    let name = gdk::keyval_name(ev.get_keyval());
+    let name2 = ev.get_keyval().name();
+
+    assert_eq!(name, Some("A".into()));
+    assert_eq!(name, name2);
+}
+
```
GuillaumeGomez, line 25: Please remove this extra line.

GuillaumeGomez:
Apart from the extra line, looks all good to me. Thanks!

tests/check_event.rs
danigm: Removed!

GuillaumeGomez: :+1:

src/enums.rs
```diff
@@ -2280,4 +2282,19 @@ pub mod key {
     pub const Prev_VMode: Key = gdk_sys::GDK_KEY_Prev_VMode as u32;
     pub const LogWindowTree: Key = gdk_sys::GDK_KEY_LogWindowTree as u32;
     pub const LogGrabInfo: Key = gdk_sys::GDK_KEY_LogGrabInfo as u32;
+
+    pub trait Keys {
+        fn to_unicode(self) -> Option<char>;
+        fn name(self) -> Option<GString>;
+    }
+
+    impl Keys for Key {
```
sdroege, line 2291: Why not `impl Key` (and `pub fn`s) here?

I think `Key` is also just a type alias: as part of this it should become a newtype wrapper (`pub struct Key(u32)`, deriving clone, copy, partialeq, eq, partialord, ord, hash at least plus impl `FromGlib` / `ToGlib`).

danigm: I created the trait because the rust compiler told me to do that when I tried to `impl Key`, but as you said I tried with the type alias, if we convert to a newtype that should work.

danigm: I've added a new commit trying to implement the functionality this way, but I'm not sure if it's correct what I'm doing, and with this new commit the API changes so any other crate that uses `gdk::enum::keys` will be broken.

sdroege:
Looks good to me but might need some fixes in gtk now. I wonder if this huge list of keys could be autogenerated so it stays in sync. Some little Rust application (or even a Python script) running on the `.gir` file shouldn't be much work.

EPashkin:
:+1:
Seems no use of `Key` type in generated code so there will be no problem with this change.

src/enums.rs
```diff
@@ -5,2279 +5,2321 @@
 #[allow(non_upper_case_globals)]
 pub mod key {
     use gdk_sys;
+    use glib::GString;
+    use keys;
+    use glib::translate::*;
 
-    pub type Key = u32;
+    #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct Key(u32);
```
sdroege, line 13:
```
    |                                                ^^^^^^ `gdk::enums::key::Key` cannot be formatted with the default formatter
```

This needs some fixing though. Can you impl the `Display` trait?

And imho all this should be in `src/key.rs`, let's not pollute the enum file with more things than necessary :)

sdroege: @danigm?

danigm: I've moved the `Key` struct to the `keys` module and leave the `enums` module as it was. I've also implemented the `Display` trait, but I don't know what to write there, I've just done the same that the `Debug` does. Maybe we can print the unicode, but we should fallback to something for non unicode keys, or the value in hexadecimal, but I'm not sure what will be the use case for the `Display` trait.

sdroege:
@GuillaumeGomez CI fails for something unrelated. Ubuntu strikes again

```
Failed to connect to Mir: Failed to connect to server socket: No such file or directory

Unable to init server: Could not connect: Connection refused

(check_event-efee6947c862df58:8815): Gdk-WARNING **: cannot open display: 
```

sdroege:
Ah that's actually a test added as part of this PR. If we can't run anything using GDK on the CI it should probably be disabled, or someone needs to figure out how to teach both travis and appveyor to run it.

src/enums.rs
sdroege: You still left the `pub type Key = u32;` and all the key constants as they are. Can you remove that and move the key constants over to the new file too while making them of the new `Key` type like you did before? Having two types for the same thing is rather suboptimal.

Apart from that this looks good to me now.

src/keys.rs
```diff
@@ -63,3 +63,2281 @@ impl ::std::fmt::Display for Key {
         write!(f, "Key({})", self.0)
     }
 }
+
+#[allow(non_upper_case_globals)]
+pub mod constants {
+    use ::keys::Key;
```
sdroege, line 69:
```suggestion
    use ::super::Key;
```

Maybe?

sdroege:
Thanks, looks good to me now. The test will still fail but maybe @GuillaumeGomez has an idea what to do about that.

GuillaumeGomez:
I just love to debug with empty logs. :)

I guess it's because of the window manager... I'll try another solution later on. Definitely don't have motivation right now.

sdroege:
Let's disable the tests for now and get it in then? @danigm @GuillaumeGomez 

(I think this just needs running the tests under Xvfb btw but I don't know how to do that with travis)

GuillaumeGomez:
I guess it's worth trying.

sdroege:
@GuillaumeGomez All green finally, good to go :)

GuillaumeGomez:
Thanks!

PR #352, opened 2020-08-09
add GDK_BUTTON_* see https://github.com/gtk-rs/gtk/issues/1044

emmanueltouzery:
I realized that I opened the bug on the wrong repo... Anyway the discussion is there https://github.com/gtk-rs/gtk/issues/1044

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -147,3 +147,12 @@ pub use window::WindowAttr;
 pub type key = i32;
 
 pub use self::keys::{keyval_name, keyval_to_unicode};
+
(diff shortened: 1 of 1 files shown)


sdroege:
@GuillaumeGomez CI is green btw :) we should probably do bugfix releases with this and the pango change

GuillaumeGomez:
It's not really a but in itself though, so not sure...

sdroege:
True but it's a trivial addition that can't possibly break anything but can make user-code better to maintain :)

PR #354, opened 2020-08-18
add missing GDK_EVENT_*

### LINKED ISSUE #336: Add GDK_EVENT_* & GDK_BUTTON_* to the public api
Per the discussion in that chat room

the gdk_button & gdk_event are exposed through the -sys crate only and should be part of the public API

Quoting ebassi

> There are a couple of defines that should be in the public API; like GDK_BUTTON_* and GDK_EVENT_PROPAGATE|STOP

bilelmoussaoui:
I wonder though if we can make things a bit better?

closes #336

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -156,3 +156,9 @@ pub const BUTTON_MIDDLE: u32 = gdk_sys::GDK_BUTTON_MIDDLE as u32;
 
 /// The secondary button. This is typically the right mouse button, or the left button in a left-handed setup.
 pub const BUTTON_SECONDARY: u32 = gdk_sys::GDK_BUTTON_SECONDARY as u32;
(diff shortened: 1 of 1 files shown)


sdroege:
Better how? We could use proper types (enums) for these things instead of `u32` inside the structs I guess, but that all has to be done manually.

bilelmoussaoui:
I was wondering if we could have something like glib::Continue but that's too much work for now

sdroege:
Agreed. That would be nicer :) Can you create an issue if you don't want to do it now?

sdroege:
@GuillaumeGomez Good to go apart from the usual mac problems with travis.

GuillaumeGomez:
Thanks!

PR #358, opened 2020-09-30
Add keyval_to/is_upper/lower functions to `Key` struct

Amjad50:
added:
- `keyval_is_upper`
- `keyval_is_lower` 
- `keyval_to_upper` 
- `keyval_to_lower`

As I could not use them in my project in the recent versions, [here]( 
https://github.com/Amjad50/plastic/blob/984590c4ede15b9ed8e028d16ea4b66a9addbfcc/nes_ui_gtk/src/ui.rs#L145), As `u32` cannot be converted back into `Key`

what do you think? @GuillaumeGomez

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -7,16 +7,6 @@ use glib::translate::*;
 use glib::GString;
 use libc::c_uint;
 
-pub fn keyval_name(keyval: u32) -> Option<GString> {
-    skip_assert_initialized!();
-    unsafe { from_glib_none(gdk_sys::gdk_keyval_name(keyval as c_uint)) }
-}
-
-pub fn keyval_to_unicode(keyval: u32) -> Option<char> {
-    skip_assert_initialized!();
-    unsafe { ::std::char::from_u32(gdk_sys::gdk_keyval_to_unicode(keyval)).filter(|x| *x != '\0') }
-}
-
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct Key(u32);
 
@@ -31,6 +21,13 @@ impl ::std::ops::Deref for Key {
 impl ::std::ops::DerefMut for Key {
     fn deref_mut(&mut self) -> &mut u32 {
         &mut self.0
+    }
+}
+
+impl ::std::convert::From<u32> for Key {
+    fn from(value: u32) -> Self {
+        skip_assert_initialized!();
+        Key(value)
     }
 }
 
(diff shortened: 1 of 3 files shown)


sdroege (approved):
Looks good to me, but maybe we should also add `From<u32> for Key` and `From<Key> for u32` impls? That's basically there already anyway via the `DerefMut` / `Deref` impls

Is there any reason why this could cause problems?

Amjad50:
oops, actually it appears I can convert without this PR:

```
let mut val = event.get_keyval();
*val = keyval_to_upper(*val);
match val {
...
```

I don't know why I struggled a lot for converting it to uppercase, that's why I suggested this PR.
Do you think this would be good anyway? I think its better for code clarity.

> we should also add `From<u32> for Key` and `From<Key> for u32` impls?

Yes, though I think `From<Key> for u32` is not needed as it is identical to `Deref`

sdroege:
>  Do you think this would be good anyway? I think its better for code clarity.

I agree, yes

sdroege (approved):

src/keys.rs
```diff
@@ -17,6 +17,26 @@ pub fn keyval_to_unicode(keyval: u32) -> Option<char> {
     unsafe { ::std::char::from_u32(gdk_sys::gdk_keyval_to_unicode(keyval)).filter(|x| *x != '\0') }
 }
 
+pub fn keyval_is_lower(keyval: u32) -> bool {
```
GuillaumeGomez, line 20: Do we really need these functions? Having them implemented on `Key` isn't enough? (Same question for the existing ones.)

Amjad50: yah, I just copied the style of the existing ones

GuillaumeGomez: Maybe we should just remove them... What do you think @sdroege ?

sdroege: I don't see the point in them either :)

Amjad50: ok I'll do it

GuillaumeGomez (approved):

sdroege (approved):

Amjad50:
I guess `keyval_name, keyval_to_unicode` are needed in `lib.rs`, they are being exported to outside

Amjad50:
since now `u32` can be converted to `Key`, are `keyval_name(u32), keyval_to_unicode(u32)` necessary?

GuillaumeGomez:
I don't think we should keep them.

Amjad50:
problems with `check_init_asserts`?

GuillaumeGomez:
You might need to add some `skip_assert_initialized` where it's missing.

